PR #4, opened 2016-12-08
Swap the position of the brackets and parentheses.

kynowoo:
The character sequence [ ] and then  () is recognized as link text by markdown so that they are eaten.


ghogen:
Thanks for your contribution!

PR #5, opened 2016-12-08
Change the markdown bold ** pairs to html bold tag

kynowoo:
It seems going wrong if /\* and \*/ was bounded by \*\*.
Suggest that you may developing a tool to check and fix these markdown conflicts when generating the documents.


ghogen:
Thanks!

PR #32, opened 2017-05-16
Align coding style

VSChawathe:
The last instantiation of call line #42 had mismatched showing filter's return value with the two previous instantiations lines #19, #30 from the same code block.


VSChawathe:
Agreed. I must mention, the reasoning to elect lookup ease from retaining the explicit return value was this being canonical documentation.

PR #39, opened 2017-06-23
Update register-usage.md

hikaru-i:
Callee cannot tell if it is needed or not.


ghogen:
@corob-msft can you take a look?

PR #41, opened 2017-06-29
Fix odd indentation in _popen/_wpopen

chwarr:
If you'd rather I not also fix the trailing whitespace errors, I can remove that commit. Let me know.

### DIFF
--- a/docs/c-runtime-library/reference/popen-wpopen.md
+++ b/docs/c-runtime-library/reference/popen-wpopen.md
@@ -69,16 +69,15 @@ Creates a pipe and executes a command.
   
 ## Syntax  
   
-```  
-  
-      FILE *_popen(  
-const char *command,  
-const char *mode   
-);  
-FILE *_wpopen(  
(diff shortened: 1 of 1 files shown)


corob-msft:
The trailing whitespace is significant in our markdown schema. If there aren't two trailing space characters after or between paragraphs, they are merged into a single paragraph in the output HTML, which is rarely what you actually want. This may not be reflected in the preview on GitHub. I suggest dropping that commit.  
  
Within triple-tick code blocks, the trailing spaces are not necessary, though they don't make a visual difference. I tend to remove them when I find them because they may be unexpected when copied into Visual Studio or another editor. Our automated conversion tools that built the markdown files often inserted them, as well as the unexpected space at the beginning of some blocks. I tend to fix those as I come across them. I did some regex searches across the entire document set to fix many, but clearly missed some. These changes are welcome.

chwarr:
Removed the whitespace commit. Thanks for the explanation.

PR #56, opened 2017-08-08
Update compile error c-3821

JackGrinningCat:
Without seeing any variable argument list and no inline assembly, i got the error

### DIFF
--- a/docs/error-messages/compiler-errors-2/compiler-error-c3821.md
+++ b/docs/error-messages/compiler-errors-2/compiler-error-c3821.md
@@ -74,3 +74,30 @@ int main() {
    }  
 } // C3821  
 ```  
+
+## Example  
+ The following sample generates C3821.  
+  
+```  
+// C3821c.cpp  ( ? )
+// compile with: /clr  
+// processor: /x64  
+#include <vcclr.h>  
+#include <atomic>
+
+class nativeInteropClass 
+{
+private:
+	gcroot<System::Object^> syncObj;
+
+	std::atomic<int> AtomicSuperCounter;
+public:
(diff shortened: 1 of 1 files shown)


corob-msft:
This looks like you're asking a question in the form of a code example in the documentation. Can you explain why this example causes C3821, or is that what you want to know?

JackGrinningCat:
Hi,Na, I never felt very assured by the error messages from the visual studio compiler. The tend to be for super genial users. So in this case it was not obvious where an error is coming from. You are right, I'm not sure, why the atomic raises such error, I just guess, the build environment v141, has new code which the version v140 has not. It might be also an error of visual studio, but I did yet find the time to pin point it.Here is it the use of the atomic variable, but the error is never marking the line of use. So I had to deconstruct my constructor, where it was occurring, so besides the wide area error message, nothing was pin pointing to the atomic. So if you tried to compile the code you will see the error, and the visual impression I saw in my code is not in this error description, so I saw it missing, as I reduced it to this simpler case, I would find it a good supplement of possible reason for this error.Well, just my good will to combine error messages with possible code patterns and the use of atomic within CPP clr code.Take it or leave it. You are Microsoft employee in the end. So you might get find the best use of this code snippet.Am 08.08.2017 10:36 nachm. schrieb Colin Robertson <notifications@github.com>:This looks like you're asking a question in the form of a code example in the documentation. Can you explain why this example causes C3821, or is that what you want to know?

ghogen:
@corob If this is not suitable for the docs, please close.

corob-msft:
After some consultation with the devs, this doesn't seem like something that needs to be called out specifically. Closing.

PR #65, opened 2017-08-27
Explain how to lookup defined macros and their value.

monty241:
I needed Google and Stackoverflow based upon this documentation to help me find the macro's values.


ghogen:
@mblome please review

corob-msft:
The instructions could be clarified, but this example is a little too specific, and thus confusing. You can get to the Macros pane from any property edit dialog, not just some specific post-build property which, as near as I can tell, does not exist in VS 2017, or at least, not as described.

PR #115, opened 2017-10-16
Pipe characters were not escaped properly.

pedro-w:
Inside backticks, the hex entities were showing up as the literal characters `&#124;`

This is just a minor typo correction to the table.


pedro-w:
I should add: I found this typo on your website, https://docs.microsoft.com/en-gb/cpp/cpp/cpp-built-in-operators-precedence-and-associativity
Formatting a pipe character in backticks in a table seems to be a bit of a 'corner case' for markdown, so your website might not convert to HTML the same way that github does it.
An alternative to backtick-backslash-pipe-backtick that works (on github, at least) is `<code>&#124;</code>`

PR #167, opened 2018-02-09
Remove __declspec(noalias) from example

manbearian:
Remove __declspec(noalias) from example to make it clearer what is going on.
I'm submitting a second PR that updates __declspec(alias) to have its own example.
Clean up other text.


corob-msft:
@manbearian Please take a look before I merge this.

PR #171, opened 2018-02-15
"Users" of the C++ compiler are "developers"

AndrewPardoe:
Users of the C++ compiler are generally developers. Isn't it a bit more precise and a bit more humanizing to refer to them as such?


AndrewPardoe:
If we had the engagement rate of products with users we’d be in way better shape.

And no, I’m not talking about software.

PR #175, opened 2018-02-19
Provide common _MSC_VER values

KindDragon:


corob-msft:
I appreciate the effort that you put into making this addition, and I agree that this info is relevant and in the right place. However, there are a couple of reasons why this kind of version-specific information is not usually added to our topics. The first is, it goes out of date rapidly. When we put a table like this in a topic, it represents an ongoing maintenance point that we have to touch with every version and update of the compiler. Our documentation already has too much out-of-date material that we are perpetually trying to find and fix. The second is that, over time, tables like these tend to get large and turn into visual clutter, filled with info for old versions that few people care about. And in cases like this one, there's a third reason, which is that the people who actually need this information can easily find it out for themselves, by following the instructions in the topic to run CL for the relevant toolsets. 

That's why I'm closing this PR unmerged. Feel free to argue the case; my opinions are not set in stone.

KindDragon:
Thanks for detailed answer

PR #193, opened 2018-03-15
Backticks, not bold.

DrPizza:


DrPizza:
Ah, the other things I looked at didn't really give me a strong sense either way, but using both right next to each other seemed particularly bad!
________________________________

PR #206, opened 2018-04-03
Inidicate that code blocks contain C++

robert-andrzejuk:


robert-andrzejuk:
This code requires more updates: 
> C++ has no built-in byte type. Use unsigned char to represent a byte value.
This is not true anymore from C++17.

robert-andrzejuk:
Do I understand correctly that comment to be tongue in cheek?

```std::byte``` is a part of C++17.

In light of Your answer then this part is not true :
>  Use unsigned char to represent a byte value.

But also please consider that this document http://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0298r3.pdf says:

>   Taken  altogether,  ```std::byte```  is  just  as  “builtin” as  if  it  was  designated  by  dedicated  keyword

corob-msft:
Ah, I must apologize, I now see that **std::byte** did make it into the final C++17 standard after all. I should have looked at that, instead of the earlier drafts. You are right, it does get mentioned in a few places, but my reading of the standard indicates that these references to **std::byte** don't require any special compiler behavior, and should work for any UDT defined the same way. It's still a library type, not built-in. And you can get it with [/std:c++17](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version) as well as **/std:c++latest**, which is handy.

PR #209, opened 2018-04-09
Add link to `basic_ifstream` documentation

andschwa:
Just like the links in the other remarks.


andschwa:
@corob-msft No problem!

PR #215, opened 2018-04-24
Updating CMake build instructions

robotdad:
Updating CMake build instructions to point to the Microsoft repo as it matches the build used in VS.


mikeblome:
Thanks!

Issue #217, opened 2018-04-29
Using uniform initialization


tuncatunc:
Hi,
It may be a good practice using uniform initialization at the following examples

- cpp-docs/docs/cpp/codesnippet/CPP/how-to-create-and-use-shared-ptr-instances_3.cpp 
- cpp-docs/docs/cpp/codesnippet/CPP/how-to-create-and-use-shared-ptr-instances_4.cpp 

```
vector<shared_ptr<Song>> v{
    make_shared<Song>("Bob Dylan", "The Times They Are A Changing"),
    make_shared<Song>("Aretha Franklin", "Bridge Over Troubled Water"),
    make_shared<Song>("Thal�a", "Entre El Mar y Una Estrella")
  };
```

instead of
```
    vector<shared_ptr<Song>> v;
    v.push_back(make_shared<Song>(L"Bob Dylan", L"The Times They Are A Changing"));
    v.push_back(make_shared<Song>(L"Aretha Franklin", L"Bridge Over Troubled Water"));
    v.push_back(make_shared<Song>(L"Thal�a", L"Entre El Mar y Una Estrella"));
```

BTW, this is my first issue on this project. 
Thanks ❤️

corob-msft:
Seems reasonable. @tuncatunc Feel free to put together a PR for this change. @mikeblome will likely review.

tuncatunc:
Pull request #220 is created.

PR #220, opened 2018-05-01
Uniform initialization is used.

tuncatunc:
Using uniform initialization for some examples makes code more readable.


corob-msft:
Fixes #217

PR #221, opened 2018-05-01
Example cannot be compiled

lyh-kernel:
There are some errors of the doc. The code cannot be compiled:

$ cat error.cpp
const int some_object = 5 ;
int volatile *const z = &some_object;

$ gcc -c error.cpp
error.cpp:2:26: error: invalid conversion from ‘const int*’ to ‘volatile int*’ [-fpermissive]
 int volatile *const z = &some_object;
                          ^


mikeblome:
Thanks @lyh-kernel! This compiles with latest MSVC so I'm following up with compiler team.

PR #222, opened 2018-05-03
Fix: -Correct texture dimension in Table "Texture Size limitation"

logmanPL:


PRMerger10:
@logmanPL : Thanks for your contribution! The author, @mikeblome, has been notified to review your proposed change.

PR #231, opened 2018-05-14
Update bit field restrictions to address #57

### LINKED ISSUE #57: Clarify "Initializing a reference with a bit field" restriction
https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/cpp/cpp-bit-fields.md

The page says:

> Restrictions on bit fields
> The following list details erroneous operations on bit fields:
> *  Taking the address of a bit field.
> *  Initializing a reference with a bit field.

The second point is not accurate.
* If it says that both references, const and non-const, are erroneous, then entire section should go to the **Microsoft Specific** section.
* Otherwise, it the page should distinguish those types of references clarifying which are supported and which are not, for bit fields, in comparison to C++ standard

http://en.cppreference.com/w/cpp/language/bit_field says
> When initializing a const reference from a bit field, a temporary is created (its type is the type of the bit field), copy initialized with the value of the bit field, and the reference is bound to that temporary.

corob-msft:
Update restriction on bit field references to say non-**const** references. Add explanatory sentence about const references binding to temporaries.

Linted markdown throughout.

Fixes #57

### DIFF
--- a/docs/cpp/cpp-bit-fields.md
+++ b/docs/cpp/cpp-bit-fields.md
@@ -12,74 +12,77 @@ ms.author: "mblome"
 ms.workload: ["cplusplus"]
 ---
 # C++ Bit Fields
-Classes and structures can contain members that occupy less storage than an integral type. These members are specified as bit fields. The syntax for bit-field *member-declarator* specification follows:  
-  
-## Syntax  
-  
-```  
-  
-declarator  : constant-expression  
-```  
-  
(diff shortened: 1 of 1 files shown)


PRMerger13:
@corob-msft : Thanks for your contribution! The author, @mikeblome, has been notified to review your proposed change.

corob-msft:
@mikeblome  Still needs a review.

PR #244, opened 2018-05-22
Removed wrong sentence about high_resolution_clock

Vizor:
Because high_resolution_clock is typedef for steady_clock as is also writen in this documentation.


PRMerger16:
@Vizor : Thanks for your contribution! The author, @corob-msft, has been notified to review your proposed change.

PR #246, opened 2018-05-22
Update compiler-error-c3533.md

CaseyCarter:
Fix number typos in the text, and clarify that the second example is a valid template definition in C++17.


PRMerger6:
@CaseyCarter : Thanks for your contribution! The author, @corob-msft, has been notified to review your proposed change.

PR #262, opened 2018-05-31
add cpp mark to code style

baruchiro:


corob-msft:
This issue was addressed in a more general PR that's since been published, so we're closing this one.

PR #298, opened 2018-06-19
Remove Incorrect /Gs Reference from /O1 and /O2

manbearian:
/O options don't set /Gs, I don't know how they got listed here (I think its been there for a while...)


PRMerger17:
@manbearian : Thanks for your contribution! The author, @corob-msft, has been notified to review your proposed change.

PR #344, opened 2018-07-13
Corrected C++/CLR to C++/CLI

svick:
Even though the switch to enable it is /clr, the language is called C++/CLI, not C++/CLR.


PRMerger13:
@svick : Thanks for your contribution! The author, @corob-msft, has been notified to review your proposed change.

PR #418, opened 2018-09-09
Update compiler-warning-level-1-c4627.md

corob-msft:
Make a more useful explanation and include an example.


PRMerger13:
@corob-msft : Thanks for your contribution! The author, @"corob-msft", has been notified to review your proposed change.

PR #421, opened 2018-09-11
RAII is not always good. Give an example of that.

snnn:


PRMerger14:
@snnn : Thanks for your contribution! The author, @"mikeblome", has been notified to review your proposed change.

PRMerger15:
@GitHubber17 : Thanks for your contribution! The author, @mikeblome, has been notified to review your proposed change.

snnn:
This is a small change. If there is no comments or suggestions, I'll abort this PR.

mikeblome:
@snnn thanks for the suggestion! You make a very good point but I don't think this topic is the best place to introduce a discussion of error codes vs exceptions.  There is another topic in that same section of the documentation that discusses how to work with code that returns error codes rather than throwing exceptions. Do you agree that this topic addresses the point you raise in this PR? https://docs.microsoft.com/en-us/cpp/cpp/how-to-interface-between-exceptional-and-non-exceptional-code?view=vs-2017 ?

snnn:
Agree.

PR #430, opened 2018-09-13
Update za-ze-disable-language-extensions.md

shsha-msft:
The removed text
"For example, /Za and /fp (Specify Floating-Point Behavior) may result in floating-point type promotion behavior that does not conform to the ISO C99 or C++11 standards."
seemed to have been added for /Za and /fp:fast. We made that combination incompatible a long time ago (2003).
In my experiments with /fp:precise and /fp:strict, the option /Za did not make any difference.


PRMerger18:
@shsha-msft : Thanks for your contribution! The author, @"corob-msft", has been notified to review your proposed change.

PR #468, opened 2018-10-15
CC74711: Wrong reference to CONTRIBUTING.md

v-makoud:
Hello, @ghogen, @Taojunshen, @corob-msft,
This proposed file change comes from https://github.com/MicrosoftDocs/cpp-docs.de-de/pull/9 .
Could you review this contribution and help to merge if agreed?
Many thanks in advance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ The documentation for Visual Basic and Visual C# are located in a separate repos
 
 ## Contributing to the documentation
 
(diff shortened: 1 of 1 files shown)


corob-msft:
@v-makoud This seems to work in GitHub without the change. Is that not true in the language-specific repos?

corob-msft:
@v-makoud It looks like the real issue is that the de-de repo does not have a localized copy of CONTRIBUTING.md, which is what the fix in MicrosoftDocs/cpp-docs.de-de#9 is trying to remedy by linking to the primary repo's English version. The right change would be to localize CONTRIBUTING, or if that's not possible, then to make a localized file-only change along the lines of the fix above.

v-maudel:
Thanks @corob-msft

PR #472, opened 2018-10-16
Remove VS2015 mentions

robotdad:
This is the very first page someone hits when they follow VS -> C++ in the documentation hierarchy. Mentioning VS2015 for legacy scenarios is a distraction from this content and shouldn't be here.


PRMerger15:
@robotdad : Thanks for your contribution! The author, @corob-msft, has been notified to review your proposed change.

corob-msft:
@robotdad I'm closing this, at least temporarily, while we work out a more general solution.

PR #511, opened 2018-10-28
typo: prefered -> preferred

nschonni:
Note: I didn't try and test to see if this is a product bug with the typo in the code vs. a docs issue


PRMerger20:
@nschonni : Thanks for your contribution! The author, @mikeblome, has been notified to review your proposed change.

corob-msft:
It's an idea. This is one of the things IntelliSense, or your build, will catch for you in case you inadvertently spell it correctly instead of as specified, so I'm not terribly worried about it. And it's definitely not getting fixed in source after more than a decade like this.

PR #515, opened 2018-10-30
Add tips section to enumerations-cpp.md

music-mind:
Add tips section to enumerations-cpp.md.

Good use case of enumerations for switch statements!

### DIFF
--- a/docs/cpp/enumerations-cpp.md
+++ b/docs/cpp/enumerations-cpp.md
@@ -175,6 +175,41 @@ int main()
 }
 ```
 
+## Tips
+
+Enumerations are useful for handling different logic cases inside of switch statements, since C++ does not recognize strings as valid switch conditions. Example:
+
+```cpp
+namespace ConditionalEnums
+{
+    enum class Suit { Diamonds, Hearts, Clubs, Spades };
+
+    void handleSuit(Suit suit)
+    {
+        switch(suit) {
+            case Suit::Diamonds: ... // Handle each case with unique logic
+            case Suit::Hearts: ...
+            case Suit::Clubs: ...
+            case Suit::Spades: ...
+        }
+        ...
+    {
+}
+```
+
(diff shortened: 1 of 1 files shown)


PRMerger15:
@music-mind : Thanks for your contribution! The author, @mikeblome, has been notified to review your proposed change.

corob-msft:
@music-mind I appreciate the effort you've put in here. Speaking generally, the language details topics are not where we discuss what a particular language feature is used for, except incidentally. That's a tutorial function, not a language reference function. In other words, we don't mention it for a reason; this isn't really the place for a "Tips" section that attempts to answer questions such as "What's an enumeration used for?"

I'm also afraid that this section doesn't meet the [voice](https://docs.microsoft.com/en-us/contribute/style-quick-start) or coding standards of our content. I'll explain a bit more, because I want to be encouraging about contributing. It's not always easy to tell what needs work.

> Enumerations are useful for handling different logic cases inside of switch statements, since C++ does not recognize strings as valid switch conditions.

The two clauses of this sentence do not have any logical relationship. Both are independently true, but the first is not a consequence of the second. When we have dependent clauses, we usually put the reason first, to make the relationship clear: "Because unscoped enumerations put unqualified names in the scope where they are declared, they can pollute the global namespace."

When we put code samples into our topics, we try to create examples that can be copied, pasted, and compiled, though sometimes with diagnostics that are specified explicitly. Old examples make use of ellipsis, but we try hard to avoid that now, and use comments instead for bits where `/* your code goes here */`. 

> To get data coarsed into the correct enum, a simple conversion function can be used to return a certain enum based on certain strings!

This sentence isn't nearly as exciting as the exclamation point would make it seem. I'm not sure what the first clause means in this context. And while using a function to match input strings to enumeration values is interesting and sometimes useful, it's not necessarily good engineering practice: it's brittle, and not localizable. 

I'm closing this PR, mostly because it's not the place for it, and only secondarily for the other issues; we're happy to work with contributors to fix that sort of thing.

PR #528, opened 2018-11-07
Update gitignore

nschonni:
Not sure if the `pfx` file should get added back, but the rest are local user files


nschonni:
No problem, feel free to fix internally

mikeblome:
ok will do thanks!

PR #536, opened 2018-11-12
Fix weird wording

manbearian:
Fix some weird wording that appears to say that the compiler will do something when we meant to say the compiler will allow the user's program do such a thing.


PRMerger18:
@manbearian : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #537, opened 2018-11-13
Removing std namespace before shuffle()

yves-dolce:
to be similar to the other <algorithm> functions documentation.


PRMerger15:
@yves-dolce : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #564, opened 2018-11-19
chore: Remove generated Markdown GIF image title

nschonni:
Used a regex `\.gif\s+"\w*"\)` to find/replace the placeholder `title` values. Since these are the same as the image file name, they don't add value


PRMerger8:
@nschonni : Thanks for your contribution! The author, @, has been notified to review your proposed change.

ktoliver:
@corob-msft - Would you review the proposed updates? Thanks.

corob-msft:
@nschonni Thanks for looking into this. Your intentions are good, but the solution is slightly off. You are right that nobody cares what the original image file name was. The source link text turns into the image's `alt` text attribute, and the text you deleted turns into the `title` attribute. However, in three out of three browsers I tried (Edge, Chrome, Firefox) the hover tooltip is populated by the `title` attribute, not the `alt` attribute (which is still needed for screen readers, I believe). The preferred solution is not just to remove the useless image file names from the titles, but to replace them with the link text (so, yes, that's also duplicated info, but it has the right effect in real world browsers). I've been doing this on a document-by-document basis as I touched them, but hadn't gone through them wholesale. We probably need the same work done for .png files, as well.

I'll close this PR, since it's not quite what I'd like to see merged. If you're willing to revisit it and run another regex on images that captures the link text and adds or replaces it as the title text, that will get a welcome reception. Thanks for all your contributions!

nschonni:
> However, in three out of three browsers I tried (Edge, Chrome, Firefox) the hover tooltip is populated by the `title` attribute

Correct, but `title` also doesn't work for keyboard and touch users. There are CSS things that can be driven to expose that info, but if the information should be visable, a figcaption is probably more appropriate

> We probably need the same work done for .png files, as well.

That was done separately in #565

> I'll close this PR, since it's not quite what I'd like to see merged. If you're willing to revisit it and run another regex on images that captures the link text and adds or replaces it as the title text, that will get a welcome reception.

Nope, I'm just going to delete this. Duplicating the alt text in the `title` isn't correct either. This was just removing the generated text that gave bad tooltips.

PR #578, opened 2018-11-28
Update fread.md to include clobbering effects in text mode

stwish-msft:
Since istream::read has started using fread(), we've received multiple reports that they expected the data past the amount read to be the same as prior to the call. This update clarifies the reasons and effects of the buffer clobbering that occurs.


PRMerger18:

PR #581, opened 2018-11-30
Add description of how to set the linker options from msbuild

bgianfo:
I'm constantly looking up how to set compiler and linker options in msbuild.
This commit starts to do that.

Not sure if you guys will be open to this type of thing or not, but I figured I would try. 😄

### DIFF
--- a/docs/build/reference/incremental-link-incrementally.md
+++ b/docs/build/reference/incremental-link-incrementally.md
@@ -66,6 +66,18 @@ Additionally, LINK performs a full link if any of the following situations occur
 
 1. Modify the **Enable Incremental Linking** property.
 
+### To set this linker option in an msbuild vcxproj
+
+Use the LinkIncremental tag in a Link item.
+
+```
+<ItemDefinitionGroup>
+  <Link>
+    <LinkIncremental>true</LinkIncremental>
+  </Link>
+</ItemDefinitionGroup>
+```
+
 ### To set this linker option programmatically
 
 1. See <xref:Microsoft.VisualStudio.VCProjectEngine.VCLinkerTool.LinkIncremental%2A>.


PRMerger17:
@bgianfo : Thanks for your contribution! The author, @, has been notified to review your proposed change.

corob-msft:
@bgianfo You have an ambition that's admirable, and I'm impressed that you'd propose something like this. If you want to take on something potentially major like the sort of change this represents, it's probably better to start with a GitHub issue, where we can discuss the changes and the philosophy behind them as well as the mechanics, rather than make a PR for just one topic, though I admit the PR is a good example that demonstrates what you have in mind.

When we consider changes of this sort, one thing we think about is just how universally applicable this change is. It's actually pretty rare for people to write .vcxproj files directly. It may surprise you (or not), but we don't actually recommend you write your own .vcxproj files, because for anything other than toy programs, they are complex and easy to get wrong, and our version-to-version project upgrade code often can't handle projects you've created yourself. 

Another consideration is the value/distraction axis. As it turns out, your addition is basically showing the textual change that occurs in the .vcxproj file when you follow the steps above to change the **Enable Incremental Linking** property in the IDE, which is how the majority of developers do it. This has some value to people (like you!) who are concerned about writing their own .vcxproj files. And of that number, virtually all know they can just bring up the project file in an editor and look at it to see the change the IDE makes. For most developers, this is a dash of mental molasses added to the topic, a bit of text that makes the mind go "Is this relevant to me?" and spend a couple of seconds evaluating it, only to decide it isn't. You don't remember this effect when you're reading, but it's real, and it's why we try to write as concisely yet accurately as we can. 

Basically, I need more convincing that this is a good idea and will be helpful enough to a large enough fraction of our audience to make it worthwhile, in comparison to the visual static it represents to the rest of our audience. We actually went through and deleted a swath of content in these topics that basically discussed how the .NET build system compares to the C++ system, because those audiences are gone. With the additions of Open Folder and CMake support in VS, we're more interested in changes applicable to these build systems that we expect to get greater adoption going forward. But I don't want to be too discouraging, either. My mind is not set in stone.

TrishaMc:
@corob-msft  Please confirm that this PR is ready for merge. Thank you.

corob-msft:
#hold-off

bgianfo:
> @bgianfo You have an ambition that's admirable, and I'm impressed that you'd propose something like this. If you want to take on something potentially major like the sort of change this represents, it's probably better to start with a GitHub issue, where we can discuss the changes and the philosophy behind them as well as the mechanics, rather than make a PR for just one topic, though I admit the PR is a good example that demonstrates what you have in mind.

Sure, I'll  keep that in mind. 

> When we consider changes of this sort, one thing we think about is just how universally applicable this change is. It's actually pretty rare for people to write .vcxproj files directly. It may surprise you (or not), but we don't actually recommend you write your own .vcxproj files, because for anything other than toy programs, they are complex and easy to get wrong, and our version-to-version project upgrade code often can't handle projects you've created yourself.

While I understand your point, in my experience there is an entirely separate reality of how msbuild and the toolsets are used by engineers out in the wild. I'm an engineer at Microsoft. There are thousands of teams internally doing complex things with msbuild which won't ever have a hope of working inside visual studio, and they really don't need to. I'm in the middle of converting a projects build system from an internal Microsoft build tool, to msbuild. It's in the order of 300+ projects. I realistically can't poke around in VS to create everyone of these projects, the reality is that my team and I have written them from scratch or used conversion tools, both of which require proper documentation for these options. I can assure you I'm not alone in this approach, as I've been part of numerous projects to do these types of conversions. 

> Another consideration is the value/distraction axis. As it turns out, your addition is basically showing the textual change that occurs in the .vcxproj file when you follow the steps above to change the **Enable Incremental Linking** property in the IDE, which is how the majority of developers do it. This has some value to people (like you!) who are concerned about writing their own .vcxproj files. And of that number, virtually all know they can just bring up the project file in an editor and look at it to see the change the IDE makes. For most developers, this is a dash of mental molasses added to the topic, a bit of text that makes the mind go "Is this relevant to me?" and spend a couple of seconds evaluating it, only to decide it isn't. You don't remember this effect when you're reading, but it's real, and it's why we try to write as concisely yet accurately as we can.

I certainly sympathize with the point about content overload, I can see it's carefully considered in the current documentation. Some points which may help see my side as a consumer of this documentation: 

 - I find it hard to believe that you think it's truly a better developer experience for someone to use a GUI to make some change, and then look at the output, then to have proper documentation as to how that task should be achieved? 

- Manual steps in an IDE are useless for code review, if I make a change to a project, and then a pull request how are my fellow developers supposed to know what a specific option does? They don't have any reasonable reference besides digging around in the MsBuild targets (and even then they might be out of luck). Certain options may be named well enough that they can guess, but that is not universally the case. 

- How do we convert to new build systems in the future? As you mentioned, CMake is becoming the de facto standard for native projects. How do I decipher what my current msbuild projects are doing so I can map it to an option or context in the CMake world?

Maybe this specific docs repository isn't the right place for this information.
Maybe we should  have better documentation for the MSBuild SDK's.
I do think this information should be available in the Microsoft docs somewhere, it's a huge time saver.

corob-msft:
@bgianfo  This information seems arguably as useful as the object model reference we provide. I'm just wondering if this is the right form for it to take. We like our code blocks to be copy and pasteable, and this example is difficult, as it shows the outer `ItemDefinitionGroup` tag that in a real project file would probably have some condition attribute, and would probably contain some sibling command option groups, and so on.

Would it be more useful to have a single page somewhere that has this info for all the linker/compiler/tool options? I'm imagining it in a table format, with the option, an XPath-like tag specification, and a list of values, something like this:

| Linker option | Project file XPath | Values |
|-|-|-|
| ... | | |
| INCREMENTAL | /Project/ItemDefinitionGroup/Link/LinkIncremental | `true` \| `false` |
| ... | | |

We could also conceivably include a single-line table like this in each option topic. Thoughts?

bgianfo:
@corob-msft Sounds like a great idea!

ktoliver:
Hi @corob-msft - Would you add a comment to show what action to take on this PR:

Merge the PR: `#sign-off`
Close the PR: `#please-close`
Wait on the PR: `#hold-off`

Thanks!

corob-msft:
Let's close this PR, since it's been sitting here as a place-holder for a work item instead.

PR #595, opened 2018-12-04
Clarify aligned allocation functions do not support block type.

stwish-msft:
Update aligned memory allocation functions to clearly state that block type is not a supported debug feature.


PRMerger18:

PR #618, opened 2018-12-13
Don't use undefined behavior for insert example

AndrewDeanMS:
Create a copy of the vector, so that first and last aren't iterators into the vector being inserted into.
Remove the redefinition of Iter that is incorrect and caused compilation failure


PRMerger17:
@AndrewDeanMS : Thanks for your contribution! The author, @, has been notified to review your proposed change.

GitHubber17:
@corob-msft - Would you review the proposed changes? Thanks

PR #646, opened 2018-12-21
Fixed typo in codecvt.md

jmittert:
codecvt should be in namespace std, not stdt


PRMerger20:
@jmittert : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #653, opened 2018-12-27
Update generic-text-mappings-in-tchar-h.md

drvoss:
_UNICODE is by default.


PRMerger18:

PR #664, opened 2019-01-01
Update how-to-convert-between-various-string-types.md

drvoss:
The example doesn't need to be seperated.


PRMerger18:

PR #666, opened 2019-01-01
Update cpp-language-reference.md

drvoss:
It would be helpful if the link is connected to the book information.


PRMerger18:

PR #671, opened 2019-01-03
Update _setmode() to clarify Unicode mode use case

stwish-msft:
_setmode() is designed to be used only with wide print functions. You can piece this together by cross referencing what Unicode Mode implies with the wprintf page, but the purpose should be called out explicitly here.


PRMerger7:

PRMerger7:

PR #678, opened 2019-01-06
Update welcome-back-to-cpp-modern-cpp.md

drvoss:
It is better to have parentheses.


PRMerger18:

PR #699, opened 2019-01-08
__[read/write/inc/add]gsqword intrinsics are not kernel specific.

neerajksingh:
The documentation for the gs: intrinsics (e.g. __readgsqword) incorrectly states that
these functions are only available in kernel mode. In fact, these functions are available
in any mode of compilation and will work for code running in any mode (kernel or user).


PRMerger17:
@neerajksingh : Thanks for your contribution! The author, @, has been notified to review your proposed change.

GitHubber17:
@corob-msft - Would you review the proposed changes? Thanks

PR #793, opened 2019-02-11
Update popcnt16-popcnt-popcnt64.md

CaseyCarter:
* "16-, 32-, or 64-byte" is an obvious typo.
* "The size of the value that the `popcnt` instruction returns is the same as the size of its argument." is a clear case of "parameter" vs. "argument" confusion. The size / type of the return value does not change to reflect the size/type of the expression passed to the function at a callsite - this sentence means "parameter." (Although honestly, this sentence adds little value - we should consider striking it.)

### DIFF
--- a/docs/intrinsics/popcnt16-popcnt-popcnt64.md
+++ b/docs/intrinsics/popcnt16-popcnt-popcnt64.md
@@ -9,7 +9,7 @@ ms.assetid: e525b236-adc8-42df-9b9b-8b7d8c245d3b
 
 **Microsoft Specific**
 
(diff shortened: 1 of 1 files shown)


PRMerger17:

mikeblome:
#sign-off

PRMerger9:
@mikeblome: I'm sorry - only the author of this article, @, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #824, opened 2019-02-28
Update linker-support-for-delay-loaded-dlls.md

MSDN-WhiteKnight:
"Statically linking with DLL" is not a correct terminology. "Implicit linking" should be used to describe linking via lib-file, such as in this topic: https://docs.microsoft.com/en-us/cpp/build/linking-an-executable-to-a-dll?view=vs-2017#implicit-linking


PRMerger14:
@MSDN-WhiteKnight : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #837, opened 2019-03-08
fix precondition on normal distribution

Lorac:
The implementation has an assert for sigma should be strictly greater than 0.0.


PRMerger17:
@Lorac : Thanks for your contribution! The author, @, has been notified to review your proposed change.

Lorac:
> sigma of zero forms a singularity.

Yes indeed.

Issue #735, opened 2019-01-23
Poor example for many reasons


dotfield:
I'd have expected Microsoft to be more useful than this, but then they seem to have abandoned C++ long ago in favour of their darling C# and the CLR version that isn't actually C++.

Aside from the stupid name of CMyException rather than StructuredException (no need for a C prefix, this isn't an MFC class and using C as a prefix to mean "class" is so 1990s and poor even then):

1. It doesn't derive from std::exception. All exceptions should derive from std::exception.
2. Exception should give more detail not just a number.
3. Why have you ignored PEXCEPTION_POINTERS. If that is supposed to get you more information about what went wrong, put it into the exception. (Clearly this goes together with point 2).
4. Your exception class is not const-correct.

I'll forgive your use of printf  but this is supposed to be C++.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a032d653-53e0-1d86-2174-1ae3a9ec8060
* Version Independent ID: 25ce27d4-cca5-c4d3-de2d-b06042e24d11
* Content: [_set_se_translator](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/set-se-translator?view=vs-2017)
* Content Source: [docs/c-runtime-library/reference/set-se-translator.md](https://github.com/Microsoft/cpp-docs/blob/master/docs/c-runtime-library/reference/set-se-translator.md)
* Product: **visual-cpp**
* GitHub Login: @corob-msft
* Microsoft Alias: **corob**

corob-msft:
@dotfield Sounds like you have a good idea of what you'd like to see in the examples. Every page in the C++ content has an edit button; feel free to propose a PR with the changes you have in mind. We're always happy to review contributions of corrections and enhancements to the docs. You'll find some useful instructions on the process in [Microsoft Docs contributor guide overview](https://docs.microsoft.com/en-us/contribute/).

mikeblome:
Looks like this issue was resolved by @MSDN-WhiteKnight -- thanks!

PR #839, opened 2019-03-11
Update set-se-translator.md

MSDN-WhiteKnight:
Improved examples according to https://github.com/MicrosoftDocs/cpp-docs/issues/735

- Make exception classes inherit std::exception

- Make exception class naming consistent in both samples

- Fixed second example not using exception number


PRMerger10:
@MSDN-WhiteKnight : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #887, opened 2019-03-27
Clarify naming confusion "_mul128" vs. "__mul128"

CaseyCarter:
The proper name is `_mul128`.

Verified with VS2019 Preview 4.4 that `__mul128` is not recognized as an intrinsic even if properly declared and `#pragma intrinsic`-ed.

### DIFF
--- a/docs/intrinsics/mul128.md
+++ b/docs/intrinsics/mul128.md
@@ -1,6 +1,6 @@
 ---
 title: "_mul128"
-ms.date: "11/04/2016"
+ms.date: "03/27/2019"
 f1_keywords: ["_mul128"]
 helpviewer_keywords: ["mul128 intrinsic", "_mul128 intrinsic"]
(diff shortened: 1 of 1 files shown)


PRMerger12:

PRMerger13:

CaseyCarter:
> @CaseyCarter : Thanks for your contribution! The author, @, has been notified to review your proposed change.

@corob-msft Who do we complain to about this broken PR bot message?

corob-msft:
@CaseyCarter That's a good question. Some of the PR bots get it right; this one doesn't, and I'm guessing it's because we have the author metadata field populated by folder in docfx.json instead of inside every article. Much easier to update in bulk that way, but harder to discover just by inspection of a file of interest. Let me ask around.

meganbradley:
@CaseyCarter @corob-msft - thanks, I've added this to the backlog for GitHub automation.

PR #909, opened 2019-04-02
In the compressed format, H represents 8 registers of 8 bytes each.

pmsjt:
So the size needs to be H * 8 * 8


PRMerger16:
@pmsjt : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #938, opened 2019-04-11
Unwind codes are actually written to memory in big-endian form.

pmsjt:
Also add info about the order of the ops in the array.


PRMerger7:
@pmsjt : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #939, opened 2019-04-11
Floats are indeed signed.

fodinabor:
Also see correct results in: https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/standard-library/is-unsigned-class.md


PRMerger19:
@fodinabor : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #957, opened 2019-04-19
Correct description of isnormal

CaseyCarter:
* Per the C Standard, `isnormal` returns a non-zero value if and only if "its argument value is normal (neither zero, subnormal, infinite, nor NaN)". This is the behavior we implement.
* Drive-by: replace occurrences of "template function" with the correct term "function template"

Note to self: close 848300 after this is merged.

### DIFF
--- a/docs/c-runtime-library/reference/isnormal.md
+++ b/docs/c-runtime-library/reference/isnormal.md
@@ -28,7 +28,7 @@ The floating-point value to test.
 
 ## Return value
 
-**isnormal** returns a nonzero value (**true** in C++ code) if the argument *x* is neither zero, subnormal, infinite, nor a NaN; otherwise **isnormal** returns 0 (**false** in C++ code).
+**isnormal** returns a nonzero value (**true** in C++ code) if the argument *x* is neither zero, subnormal, infinite, nor a NaN. Otherwise, **isnormal** returns 0 (**false** in C++ code).
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


PRMerger7:

PRMerger12:

CaseyCarter:
@corob-msft When/how do changes here flow to docs.microsoft.com? I made this change in response to a bug report on Developer Community, and I like to resolve the bug, but I'm hesitant to do so while https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/isnormal?view=vs-2019 still shows the old docs.

corob-msft:
@CaseyCarter 

Short answer: it should be live **Monday before noon** local time. I got to this one a little too late for today's publishing cut-off.

Long answer is, it's complicated. Changes that are reviewed and merged to the public repo are picked up and automatically moved to a new PR in the private repo, where a bunch of automated build checks run. Those checks have to pass, and if the PR changes are non-trivial, it must also be approved by a reviewer. Then it's merged. Once per business day, the merged content in the private repo is automatically built and published to Docs, usually sometime between 10 and 11 AM local time. Once the build is complete, another automated process gathers all the new commits in the private repo and makes a PR with them in the public repo, so everything stays in sync.

Since you have access to the private repo, you can skip a step in this process in the future by making your PRs there. That has the benefit of showing you the results of the build checks, plus a private review build of your content so you can see how it will look on Docs. The downside is you have to create a fork of the repo to work in. Help getting set up to work in the private repo is available in [Set up Git repository locally for documentation](https://review.docs.microsoft.com/en-us/help/contribute/contribute-get-started-setup-local?branch=master).

For truly urgent issues (live site breaks, landing page 404s, etc.) we can push content live out-of-band. It takes ten to thirty minutes to get built and published live, depending on build server load.

CaseyCarter:
Wow! Thank you for the incredibly detailed response.

PR #969, opened 2019-04-25
HFA and HVA Return Value ARM64 ABI documentation

prjoseph:
HFA and HVA values are returned in s0-s3, d0-d3 or v0-v3 registers depending on the size of the individual members of the HFA/HVA. For example a 4 member double type HFA is returned in d0-d3 with the first member returned in d0 and the 4th member returned in d3.


PRMerger17:
@prjoseph : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #985, opened 2019-05-02
ODBC and OLE DB reference updates

David-Engel:
OLE DB was undeprecated and is now under continuous development. Updating to point to the docs for the new OLE DB driver and removing the deprecation content. Also updating the content to more accurately describe OLE DB options.
For reference, see https://docs.microsoft.com/en-us/sql/connect/connect-history?view=sql-server-2017


PRMerger17:
@David-Engel : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #1093, opened 2019-06-15
?! what are you talking about

EmilEnchev:
For me in such way is more clear.

"Returns **true** if the condition is detected at each element in the indicated range or if the range is empty, and **false** otherwise."

Please add "or if the range is empty" at lease. Because if range is empty all_of return **true** too. 

P.s. I'm new in GitHub. Please if I'm wrong in some my actions on that how I do things here, tell me.


PRMerger8:
@EmilEnchev : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1326, opened 2019-07-12
removing inapropriate 32-bit specifications

dragon4eg:
CListBox::GetItemData()/SetItemData works also with 64bit values, which is tested at least in SDK 10.0.16299.0 in 64bit app.
So one can safely pass in a 64bit pointer.
Moreover functions' signatures actually imply this fact as _PTR-sized types are used.


PRMerger20:
@dragon4eg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger20:

PR #1328, opened 2019-07-12
Update /std:latest to /std:c++latest

scaryrawr:
/std:latest is an unknown option to the compiler and ends up ignored.


PRMerger14:
@scaryrawr : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1338, opened 2019-07-16
Remove mention of AddressSanitizer support under "Install the C++ Lin…

esweet431:
…ux workload in Visual Studio"

ASan support isn't very relevant when a user is trying to modify their VS install - I think this detail makes the page more confusing than it needs to be.


PRMerger17:

PRMerger17:
@mikeblome : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1339, opened 2019-07-16
Remove mention of native WSL support + separation of build/deploy

esweet431:
Remote mention of both native WSL support and separation of build/deploy from "Install the C++ Linux workload in Visual Studio." Neither are very relevant when the user is trying to modify their VS installation and add extra complexity (especially when both features are documented elsewhere)


PRMerger19:

PR #1404, opened 2019-07-23
Updated function signature.

isaac868:
OnDeleteItem function signature should have an int as it's first parameter.


PRMerger19:
@isaac868 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1451, opened 2019-07-25
Stringizing operator code example was misplaced

dpmm99:
Previously, the two descriptions were before both code blocks, but the second code block corresponds with the first description.


PRMerger17:
@dpmm99 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1476, opened 2019-08-05
Fix typo

saschanaz:
The example actually uses `remove_copy_if` so fixed it accordingly.


PRMerger12:
@saschanaz : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1493, opened 2019-08-10
Complete sentences with omitted words

lsaether:
These words were implied by the sentences but seemingly omitted.


PRMerger6:
@lsaether : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1527, opened 2019-08-27
Update #error docs

NextTurn:
The Syntax sections in this directory are quite inconsistent.


NextTurn:
Thanks for clarification @corob-msft. I agree that formatted blocks help distinguish different elements.

PR #1565, opened 2019-09-14
Updated parameters name.

goh-chunlin:
The two parameters `hwnd` and `uMsg` are not used in the subsequent sample codes and paragraphs. Instead, the names `hWnd` and `message` are used. Hence, to avoid confusion and standardize the naming in this tutorial, we need to update the names.


PRMerger7:
@goh-chunlin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1597, opened 2019-10-01
Remove what's new in 2019?

esweet431:
With Overview Pages landing in 16.4, this piece of documentation will be the first piece of external documentation that we link to to get started with CMake. I'm not sure it makes sense to start with a list of everything new in 2019 - many users are not familiar with our old support, and listing new features that they may not understand (e.g. open an existing cache, the relationship between msbuild.exe and CMakeLists.txt) could add complexity. I think it would be better to remove this section.
That said, you are the domain expert and if this is standard practice, then please let me know.


PRMerger16:

PR #1605, opened 2019-10-02
Conversions from floating point types vc2019

jpmorgan-atMS:
Correct numerous obsolete or inaccurate statements about numeric conversions, add **long long** types, and add explanation for the results of converting FP values to integer type.

The principal changes are for "Conversions from floating-point types". In particular, conversions to **unsigned long**, **long long**, or **unsigned long long** have worked correctly for many releases of Visual Studio. Because AVX-512 introduced conversion instructions for unsigned integers, VS2019 revised the behavior of conversion to **unsigned long** and **unsigned long long** on x86 to match. The Intel Architecture (IA) floating-point to integer conversion instructions return sentinel values as indicated. (ARM floating-point to integer conversion instructions return saturated values.)

"Conversions from Signed Integral Types" was clarified to indicate that when converting to a signed or unsigned integer type, the source value is sign-extended or truncated to the size of the result, and the resulting bits are interpreted using the destination type. For conversion to floating-point I replaced the ambiguous (and somewhat incorrect) statement that some precision may be lost when converting to **float**, with the more specific statement that if the result cannot be represented exactly the result will be the next higher or lower representable value. I also added a reference to Storage of Basic Types for size information. I removed a redundant reference to the name of the table as it should not be ambiguous.

"Conversions from Unsigned Integral Types" was similarly clarified and reconciled with the signed integer conversion page. The example code as shown is incorrect; I provided a correction. I am not certain that this is the best illustration of the point, but at least the revised version should work correctly. I removed the inaccurate statement that conversions to **float** are not accurate if the value being converted is larger than the maximum positive signed value. Conversions to floating-point have been accurate with any unsigned value for quite a while.

"Storage of Basic Types" was reconciled so that the names of all types are in bold, rather than some being in bold and others in monospaced typeface.

"Standard Conversions" was changed to say that truncation means rounding toward zero, rather than saying no rounding takes place. A statement about out-of-range values was added to make the description more complete.

### DIFF
--- a/docs/c-language/assignment-conversions.md
+++ b/docs/c-language/assignment-conversions.md
@@ -1,25 +1,25 @@
 ---
-title: "Assignment Conversions"
+title: "Assignment conversions"
 ms.date: "11/04/2016"
 helpviewer_keywords: ["conversions, assignment", "assignment conversions"]
 ms.assetid: 4ee01013-de32-4aae-b12e-0051d0cde927
 ---
-# Assignment Conversions
+# Assignment conversions
 
 In assignment operations, the type of the value being assigned is converted to the type of the variable that receives the assignment. C allows conversions by assignment between integral and floating types, even if information is lost in the conversion. The conversion method used depends on the types involved in the assignment, as described in [Usual Arithmetic Conversions](../c-language/usual-arithmetic-conversions.md) and in the following sections:
 
-- [Conversions from Signed Integral Types](../c-language/conversions-from-signed-integral-types.md)
+- [Conversions from signed integral types](../c-language/conversions-from-signed-integral-types.md)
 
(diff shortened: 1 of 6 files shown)


PRMerger8:
@jpmorgan-atMS : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
@mikeblome - can you take a look? 

Note that even if you approve the changes in this PR, they are too extensive to review in the public repo and they'll need to be either resubmitted to the private repository `cpp-docs-pr` or we'll need to move the commits.

corob-msft:
@Jak-MS @mikeblome @jpmorgan-atMS 
I'm looking into moving the commits now.

PRMerger20:

corob-msft:
Closing in favor of MicrosoftDocs/cpp-docs-pr/pull/2382 for reviews and automation.

PR #1606, opened 2019-10-03
Update required dependencies + small edits

esweet431:
I added a section on "required dependencies" before the distro-specific commands to inform customers why each dependency is required, to specify that Clang is available (instead of GCC) in 2019, and to link to our statically linked CMake binaries for Linux. I also made a few edits to the first few sections: 
- removed mention of profiling b/c not all profiling tools work in Open Folder
- removed some of the "what's new in 2019 content" that doesn't have to do with installing the workload or getting started - this page is linked to from Overview Pages and I think it may provide unnecessary complexity
- I also added rsync as a dependency for remote systems


PRMerger19:

PRMerger14:

PR #1608, opened 2019-10-03
update intellisense for wsl

esweet431:
remote headers still applies to wsl (it is only the build tree that is not copied over, the linux filesystem can see into windows but currently windows cannot see into wsl, so we need to rsync over the remote headers)


PRMerger19:

PR #1612, opened 2019-10-07
Fix the encoding of the MSFT_OP_* unwind codes in ARM64.

pmsjt:


PRMerger14:
@pmsjt : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1646, opened 2019-10-22
Update and rename ThirdPartyNotices to ThirdPartyNotices.md

Youssef1313:


corob-msft:
@Youssef1313 

This file can't be changed without clearance from our legal department. Other Docs repos have accepted this change, but they didn't check whether it's a file users can change. However, both of the changes made here seem quite reasonable. I'll propose these changes to the official sources, then all the Docs repos will get the change. Thanks for contributing!

corob-msft:
MicrosoftDocs/DocsLegal-pr/pull/149 for tracking purposes

Issue #1650, opened 2019-10-24
Short description is wrong/misleading


angusgraham:
"Tests whether the type has a move constructor." 
That is not true. 
As stated on https://en.cppreference.com/w/cpp/types/is_move_constructible:
Types without a move constructor, but with a copy constructor that accepts const T& arguments, satisfy std::is_move_constructible.
Short description should be changed to:
"Tests whether the type can be constructed by using a move operation"
And the Remarks should be changed to 
"A type predicate that evaluates to true if the type T can be constructed by using a move operation. This predicate is equivalent to is_constructible&lt;T, T&&&gt;.
Note: Types without a move constructor, but with a copy constructor that accepts const T& arguments, satisfy std::is_move_constructible. "

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1247c7b0-54ab-62ef-aaa3-8a457f4b5d6d
* Version Independent ID: 9d2e74fb-54db-920e-ad57-eb14ca3d9474
* Content: [is_move_constructible Class](https://docs.microsoft.com/en-ca/cpp/standard-library/is-move-constructible-class?view=vs-2019#feedback)
* Content Source: [docs/standard-library/is-move-constructible-class.md](https://github.com/Microsoft/cpp-docs/blob/master/docs/standard-library/is-move-constructible-class.md)
* Product: **visual-cpp**
* Technology: **cpp-standard-libraries**
* GitHub Login: @corob-msft
* Microsoft Alias: **corob**

corob-msft:
@angusgraham 

Thanks for pointing that out; you are correct. You're welcome to go one step further and create a PR with those changes, if you like. We have a [contributor guide](https://docs.microsoft.com/en-us/contribute/#quick-edits-to-existing-documents) with instructions. You don't have to, of course; we'll get around to it, but it might be a while.

angusgraham:
Sure thing. Made that pull request: https://github.com/MicrosoftDocs/cpp-docs/pull/1652

angusgraham:
Sure thing. Made that pull request: #1652

corob-msft:
@angusgraham 
Awesome work, thanks!

PR #1652, opened 2019-10-24
Short description is wrong/misleading

angusgraham:
As discussed in https://github.com/MicrosoftDocs/cpp-docs/issues/1650
"Tests whether the type has a move constructor."
That is not true.
As stated on https://en.cppreference.com/w/cpp/types/is_move_constructible:
Types without a move constructor, but with a copy constructor that accepts const T& arguments, satisfy std::is_move_constructible.


PRMerger6:
@angusgraham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger9:

PR #1672, opened 2019-10-29
PG0165 can also be shown when PGD is readonly

ajbarb:


PRMerger8:
@ajbarb : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@ajbarb 

As near as I can tell, the reported message number may be the same, but the text is something like:

> Reading 'filename.pgd' : 'PGD file is read-only'.

There are other possibilities, as well.

PRMerger10:

PR #1674, opened 2019-10-29
Give linux media standardized filenames

corob-msft:
Change names in PR #1673 to match doc styleguide standards.


corob-msft:
@mikeblome Mind giving this an approval?
If we change the filenames here before the others are merged into the private repo, it saves some grief.

corob-msft:
@esweet431 FYI in case you're writing content to use these filenames.

Jak-MS:
@corob-msft - changing file names in the public repo can be tricky because we don't have validation tools to check affected docs that link to the files. and if the files are already in master they'll need redirects associated, which we can't test here. (recommend moving this to the private repo).

mikeblome:
@jak-MS these are new files with no current links in the repo

Jak-MS:
@mikeblome @corob-msft - ignore my comment about redirects... i see these are image files.

PR #1694, opened 2019-11-09
Update algorithm.md

stanleylalanne:
Changed: - The `_if` suffix indicates that the algorithm is used with function objects operating on the values of the elements rather than on the values of the elements themselves.
To: - The `_if` suffix indicates that the algorithm is used with function objects operating on the values of the elements rather than on the elements themselves.


PRMerger8:
@stanleylalanne : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1733, opened 2019-11-21
Fix wrong path to the MathLibrary.lib file

hongkuancn:
If the DLL and client projects are in the same solution, `..\` goes to the solution folder, while `..\MathLibrary` goes to the project folder. `MathLibrary.lib` file should be in the solution folder instead of project folder. I have tested it with real code.


PRMerger17:
@hongkuancn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Hi @corob-msft - Could you review the proposed change? Thanks.

PRMerger15:

PR #1739, opened 2019-11-25
small grammatical fix

cuchalin:
Changed a sentence from: In practice, however, it are usually represented as a table for the first 256 characters and as ranges for the double-byte values
To: In practice, however, it is usually represented as a table for the first 256 characters and as ranges for the double-byte values


PRMerger6:
@cuchalin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1745, opened 2019-11-26
Update warning.md

DGilliesMS:
The pragma warning will not be honored at the open curly but rather must appear before the function declaration.


PRMerger18:
@DGilliesMS : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1755, opened 2019-12-02
Add language identifiers

NextTurn:


ktoliver:
Verified.

@NextTurn - Thanks for submitting the updates. Note that smaller pull requests are easier to review. In diff view, after a few hundred files (exact number unknown--appears to be around 300), all subsequent files are collapsed and must be manually opened to view diffs:

image

Just a tip that would make things easier on this end. :) Thanks again.

PR #1778, opened 2019-12-11
Add remark about _aligned_malloc and free

petrutlucian94:
The docs state that using "free" for a pointer returned by
_aligned_malloc is illegal but only as an inline comment within the
code sample.

This isn't visible enough as most people (including me) tend to skip
the code sample section. Such an oversight will lead to crashes that
can sometimes be time consuming to debug.

For this reason, this patch adds another note to the "Remarks"
section.


corob-msft:
Of course, maybe I should have waited. I'll do the other clean-up later. Thanks again.

PR #1789, opened 2019-12-16
Add language identifiers

NextTurn:


corob-msft:
@NextTurn 

I need to think about these changes before you go making them wholesale across hundreds of entries. We haven't consistently marked the library Syntax sections as C or C++ because they aren't always valid declarations. Sometimes they're just informative, with a non-specific type placeholder, or even a bit of hand-waving description. Even calling these blocks Syntax is something of a problem, since that implies a description of how they're used, not how they're declared. And despite repeated search passes, I still regularly find some are badly formatted, to say nothing of out of date. Once upon a time we could combine code formatting with italics to indicate placeholders or descriptions. We don't have that in markdown. We have... comments. Let me get a local consensus on the approach to take, then get back to you. Thanks for proactively taking things like this on.

corob-msft:
@NextTurn 
Thanks for helping out with repo consistency issues. I looked into the code blocks issue some more, and where it makes sense, I think it's okay to add a cpp tag. That requires actually looking at each untagged block for context, but anything that's strictly a declaration is okay. I tried replicating the change you made, by using a regex, and started with updating the same ones you fixed here. So, these fixes should now be in the repo, and rather than cause a merge issue, I'll close this PR. But it's not a rejection of the idea, it's just about these particular files, which have gotten pretty stale. One request: For ease of reviewing, it would be nice to limit changes to 50 files at a time or less. Thanks!

PR #1822, opened 2020-01-02
Update using-cpp-interop-implicit-pinvoke.md

jkotas:
Incorporate feedback from https://github.com/dotnet/coreclr/issues/659#issuecomment-558914602


corob-msft:
@jkotas Thanks for the update.

PR #1845, opened 2020-01-14
Add new properties for Linux IntelliSense

esweet431:
Add 3 new properties for Linux IntelliSense, one property that was added for the separation of build/deploy


ktoliver:
Hi @mikeblome - Could you review the proposed changes to the article? Thanks.

esweet431:
CC @corob-msft

PR #1886, opened 2020-01-30
Add F1 link from Command Line in C++ properties

benmcmorran:
In a C++ project, go to project properties, and select Configuration Properties > C/C++ > Command Line. Press F1. You end up on the [Clang Linker properties (Android C++) page](https://docs.microsoft.com/en-us/visualstudio/cross-platform/clanglink-prop-page?f1url=https://msdn.microsoft.com/query/dev16.query?appId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk%28vc.project.AdditionalOptionsPage%29%26rd%3Dtrue&view=vs-2019), but this page seems like a much better fit.

### DIFF
--- a/docs/build/reference/compiler-command-line-syntax.md
+++ b/docs/build/reference/compiler-command-line-syntax.md
@@ -1,6 +1,7 @@
 ---
 title: "MSVC Compiler Command-Line Syntax"
 ms.date: "11/04/2016"
+f1_keywords: ["VC.Project.AdditionalOptionsPage"]
 helpviewer_keywords: ["syntax, CL compiler command line", "cl.exe compiler, command-line syntax"]
 ms.assetid: acba2c1c-0803-4a3a-af25-63e849b930a2
 ---


PRMerger13:
@benmcmorran : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@benmcmorran 

I'm sorry to report that this change alone is unlikely to have the desired effect, or possibly any effect at all. By my count, this would make a *seventh* doc page into a target for this particular F1 tag. As near as I can tell, there's some randomness in which page gets chosen as the actual F1 redirection target when more than one has the tag. [Clang Linker Properties (Android C++)](https://docs.microsoft.com/en-us/visualstudio/cross-platform/clanglink-prop-page?view=vs-2019) happened to be the winner. The other doc pages with this tag are:

- [Command Line Property Pages](https://docs.microsoft.com/en-us/cpp/build/reference/command-line-property-pages?view=vs-2019)
- [Manifest Tool Property Pages](https://docs.microsoft.com/en-us/cpp/build/reference/manifest-tool-property-pages?view=vs-2019)
- [MIDL Property Pages](https://docs.microsoft.com/en-us/cpp/build/reference/midl-property-pages?view=vs-2019)
- [Resources property page](https://docs.microsoft.com/en-us/cpp/build/reference/resources-property-pages?view=vs-2019)
- [Clang Project Properties (Android C++)](https://docs.microsoft.com/en-us/visualstudio/cross-platform/clang-android-prop-page?view=vs-2019)

The real problem is in Visual Studio. Lots of controls are using "VC.Project.AdditionalOptionsPage" (with varying capitalization) as their F1Keyword attribute. I found no less than 27 xml files with this F1Keyword attribute in my installed VS, under MSBuild\\Microsoft\\VC\\v160. Unfortunately, I don't know where to look for the sources to edit them. Ion might. What needs to happen is to change the F1Keyword attributes in whatever xml files exist for each of those property pages' controls to something unique (at least, unique per property page), and then update each of these Docs pages' f1_keywords header to match the new unique values. Ideally the change would go out in a servicing update for older binaries, as well, though it's unlikely to meet the bug bar, since we've lived with it forever, apparently.

benmcmorran:
Thanks. I'll see what I can change on the VS side and open a followup PR once that's resolved.

PR #1893, opened 2020-02-03
Add extra F1 keyword to C++ Command Line article

benmcmorran:
This is preparation for a change in VS that will fix the F1 link from Configuration Properties > C/C++ > Command Line in Project Properties.

@corob-msft This is a more narrow fix than addressing all the vc.project.AdditionalOptionsPage issues, but will address the current feedback we have.


PRMerger20:
@benmcmorran : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1990, opened 2020-03-15
Fix typo

eltociear:
massage→message


corob-msft:
@eltociear Thanks for the fix!

PR #1994, opened 2020-03-16
fix: MD046/code-block-style

nschonni:
Some section incorrectly flagged as indented code block style


nschonni:
@corob-msft no problem, I'm going to send a PR with an updated .markdownlint.json. I've just been using the new `--fix` for some of these PRs.

PR #2008, opened 2020-03-20
Add dependency on ninja-build

esweet431:
In VS 2019 16.6 and later Ninja will be the default generator for new Linux/WSL configurations.


PRMerger18:

PR #2014, opened 2020-03-21
Update cpp-conformance-improvements.md to reference P1423R3 instead of P1423R0.

tahonermann:
P1423R3 is the revision that was adopted for C++20.


PRMerger12:
@tahonermann : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2056, opened 2020-04-03
Fix push_back prototypes

KomarDL:
"void push_back(void push_back(Type&& val);" should be 
"void push_back(const Type& val); 
void push_back(Type&& val);"


PRMerger6:
@KomarDL : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2059, opened 2020-04-04
Update walkthrough-deploying-your-program-cpp.md

bergano65:
The steps in this section vary depending on which version of Visual Studio you have installed. Make sure the version selector in the upper left of this page is set correctly.

 which upper right corner you mean? Command "Help/About"?,
 if yes better write directly

### DIFF
--- a/docs/ide/walkthrough-deploying-your-program-cpp.md
+++ b/docs/ide/walkthrough-deploying-your-program-cpp.md
@@ -20,6 +20,9 @@ The walkthrough shows how to use Windows Installer to deploy your application. Y
 
 ## Install the Visual Studio setup and deployment project template
 
+---
+???
+---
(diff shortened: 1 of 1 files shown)


PRMerger15:
@bergano65 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@bergano65 
Though submitted as a PR, it seems that you're raising an [issue](https://github.com/MicrosoftDocs/cpp-docs/issues) rather than providing a solution. The [feedback control on the page in question](https://docs.microsoft.com/en-us/cpp/ide/walkthrough-deploying-your-program-cpp?view=vs-2019#feedback) is the easiest way to bring issues to our attention, and has better tracking and reporting options. Thanks for letting us know.

PR #2113, opened 2020-04-20
fix: MD040/fenced-code-language

nschonni:
Find: ```\n(.*) void
Replace: ```cpp\n$1 void


corob-msft:
@nschonni 
An interesting approach, but I'm not convinced it's best. I'd much rather this issue get fixed by file, by directory, in small batches (< 50 files at a time), so a given file isn't updated for MD040 more than once. It takes me a long time to look at every change, especially when the numbers get high enough that GitHub doesn't expand them all for review by default, and I don't really have the time to spare. I've found a fair number of instances of things in triple-tick blocks that looked like code but weren't really, so the mass regex fixes are problematic. *Доверя́й, но проверя́й*.

nschonni:
Sorry, but you have the MD040 issue in more than 1K files, so I don't think you're going to be able to do it manually. I'll stop submitting these PRs, but this is the same thing as the PR you merged earlier today

PR #2196, opened 2020-05-15
Update move-constructors-and-move-assignment-operators-cpp.md

CaseyCarter:
Annotate the sample `MemoryBlock` move constructor and move assignment operator with `noexcept` so they correctly produce the expected behavior in the example which reallocates a vector of `MemoryBlock`.

Fixes VSO-1120480.

Drive-by: Cleanup handwavy sentence about "preserving the rvalue property".


corob-msft:
@CaseyCarter  It's in the publishing pipeline, and should be live Friday by 11AM Redmond time.

CaseyCarter:
Thanks, @corob-msft!

PR #2246, opened 2020-06-04
Fix misnamed function name: fesetenv -> fesetround

stwish-msft:
In my latest PR and previous change to Link Options, I said fesetenv changes the rounding mode, when it is actually fesetround.


PRMerger9:
@stwish-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2263, opened 2020-06-13
Meaning of the postfix-expression

yecril71pl:
Fix the following problems with the current text:
1. An expression does not identify anything, it is an identifier’s job.
2. There is no such thing as a function object.
3. A reference to a callable object need not be an lvalue expression.
4. Joining some elements of an unordered list with **or** is confusing, especially when all alternatives are of equal status and importance.

I also have also included several remarks that I consider would improve the reliability of the text.

## TO DO

1. The examples are inconsistent.
1. Links to detailed explanations of key concepts.

### DIFF
--- a/docs/cpp/function-call-operator-parens.md
+++ b/docs/cpp/function-call-operator-parens.md
@@ -7,7 +7,7 @@ no-loc: [ opt ]
 ---
 # Function Call Operator: ()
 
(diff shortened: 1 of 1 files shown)


PRMerger20:

PRMerger17:

PR #2275, opened 2020-06-22
Explain Linux target locale management

lukka:
Clarify that the remote Linux targets locale management is not handled by VS, rather it is up to the user to set the desired one.


PRMerger20:
@lukka : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2300, opened 2020-07-08
says "you must also specify \Og" but that is not true

angusgraham:
Doc for /Ot /Os says

`If you use **/Os** or **/Ot**, then you must also specify [/Og](og-global-optimizations.md) to optimize the code.`

...but if you look at the doc for /Og it says

"Deprecated. Provides local and global optimizations, automatic-register allocation, and loop optimization. We recommend you use either /O1 (Minimize Size) or /O2 (Maximize Speed) instead."

...and in the remarks section

"/Og is deprecated. These optimizations are now generally enabled by default. For more information on optimizations, see /O1, /O2 (Minimize Size, Maximize Speed) or /Ox (Enable Most Speed Optimizations)."

...so I've removed the misleading instruction to turn on /Og (which is actually not even possible through the Visual Studio options optimization pane - which makes sense since it's deprecated) and instead added a more generic message about turning on optimizations (copied the "For more information" part from https://github.com/MicrosoftDocs/cpp-docs/edit/master/docs/build/reference/og-global-optimizations.md)

Since it's no longer a direct instruction but just further info I've changed it to a [!NOTE]


PRMerger18:
@angusgraham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

angusgraham:
Looking at a grep of cpp-docs for /Og (https://github.com/MicrosoftDocs/cpp-docs/search?q=Og&unscoped_q=Og) it looks like most references to /Og are similarly misleading. I'm not going to fix them all but I suggest someone should.

PRMerger13:

PR #2331, opened 2020-07-28
Update compiler-error-c3028.md to ensure correct syntax highlighting in browser for MyClass::x_private inside OpenMP pragma

tangogu:
Ensure MyClass::x_private is in the same line as #pragma, to get correct syntax highlighting. Also removed x/g_i since previous OpenMP pragma line already covered them. This is to ensure the line width does not increase too much since previous issue may be caused by auto format post processing to ensure better experience in all devices.


PRMerger6:
@tangogu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2366, opened 2020-08-15
Update msbuild-visual-cpp-overview.md

OdinTemple:
The path is incorrect in case of vs 2019


PRMerger7:
@OdinTemple : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2497, opened 2020-09-28
Fix CRect::CRect description

matrohin:
According to the source code (see atltypes.h file) it actually initializes all members with `= 0`.

I checked that CPoint and CSize documentation is correct.
Please correct me if I'm wrong, here is the source code of CRect::CRect on my machine:
```
// CRect
inline CRect::CRect() throw()
{
	left = 0;
	top = 0;
	right = 0;
	bottom = 0;
}
```

### DIFF
--- a/docs/atl-mfc-shared/reference/crect-class.md
+++ b/docs/atl-mfc-shared/reference/crect-class.md
@@ -284,15 +284,15 @@ Specifies the bottom-right position of `CRect`.
 
 ### Remarks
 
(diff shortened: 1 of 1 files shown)


PRMerger19:
@matrohin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft (approved):
@matrohin 
Thanks for the update. It's possible the old documentation was true when it was written, but it's been incorrect for a long time.

PR #2508, opened 2020-10-01
There are no short integer literals

yecril71pl:
I replaced literals for _short_ integer types with literals for _medium_ integer types because there are no such literals (`0S` is not a valid short integer).

### DIFF
--- a/docs/cpp/numeric-boolean-and-pointer-literals-cpp.md
+++ b/docs/cpp/numeric-boolean-and-pointer-literals-cpp.md
@@ -22,7 +22,7 @@ Sometimes it's important to tell the compiler how to interpret a literal, or wha
 
 ## Integer literals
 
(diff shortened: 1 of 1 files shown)


PRMerger10:

PRMerger10:

corob-msft (approved):
@yecril71pl 
Thanks for the update. I don't know where the short part came from, unless by some analogy with `float`. I swapped "long long" in for "medium" since I'd rather not explain what the word "medium" means when it isn't used anywhere else.

yecril71pl:
> @yecril71pl
> Thanks for the update. I don't know where the short part came from, unless by some analogy with `float`. I swapped "long long" in for "medium" since I'd rather not explain what the word "medium" means when it isn't used anywhere else.

I think it goes back to the time of 16–bit CPUs.

PR #2514, opened 2020-10-03
Fix printing streampos to ostream

yecril71pl:
The expression `cout << y` works as described only if no shift operator of `streampos` into `ostream` is defined.  That operator may be defined by user code and the result may be different.

### DIFF
--- a/docs/standard-library/ios-typedefs.md
+++ b/docs/standard-library/ios-typedefs.md
@@ -61,7 +61,7 @@ int main( )
    ofstream x( "iostream.txt" );
    x << "testing";
    streampos y = x.tellp( );
-   cout << y << endl;
+   cout << + y << endl;
 }
 ```
 


PRMerger16:

TylerMSFT:
Hi Chris,
That's true, but I think it's apparent. And it complicates the mainline case. I'd rather not make this change.

yecril71pl:
> Hi Chris,
> That's true, but I think it's apparent. And it complicates the mainline case. I'd rather not make this change.

For me the mainline case is that being able to say `cout << y` means it is also possible to say `cin >> y`, and if that is not possible, it is worth to indicate this fact somehow.

docs/standard-library/ios-typedefs.md
```diff
@@ -61,7 +61,7 @@ int main( )
    ofstream x( "iostream.txt" );
    x << "testing";
    streampos y = x.tellp( );
-   cout << y << endl;
+   cout << + y << endl;
```
CaseyCarter, line 64: I don't think this example should take advantage of the fact our implementation of `fpos` happens to be implicitly convertible to `streamoff`, since the standard only requires that it can be explicitly converted (http://eel.is/c++draft/tab:fpos.operations#row-7). I also think that using unary `+` here to force that conversion is fairly arcane; it makes the example less clear rather than moreso.
Why not simply:
```suggestion
   cout << streamoff(y) << '\n';
```
(I replaced `endl` as a drive-by because we shouldn't teach it as the "default" way to end lines.)

CaseyCarter (changes requested):

TylerMSFT:
I've made this change to the topic and it will be live my Monday.

TylerMSFT:
Closing because I made the updates from Casey in another branch.

yecril71pl:
> I've made this change to the topic and it will be live by Monday.

It is Monday and it is not live.

TylerMSFT:
@yecril71pl - I checked and it's live. Perhaps try refreshing the page in your browser.

corob-msft:
@yecril71pl Publishing usually happens between 10 and 11 AM Redmond time (so, usually by 1900 UTC) on weekdays.

PR #2587, opened 2020-11-09
Explicitly convert streampos to streamoff

yecril71pl:
1. Explicitly convert `streampos` to `streamoff`.
2. Print `'\n'` instead of `endl`.


PRMerger14:

TylerMSFT (approved):
Thanks for catching the second one. The first change is already live.

PR #2613, opened 2020-11-21
Update afx-extension-module-structure_1.cpp - inaccurate initialisation of AFX_EXTENSION_MODULE structure

yury-fedorov:
The first two parameters of the structure AFX_EXTENSION_MODULE are BOOL and HMODULE. So initialisation with = { NULL, NULL } is inaccurate: first NULL is FALSE (or 0 due to BOOL is defined as int). In fact this expression cannot be transformed to modern C++ equivalent = { nullptr, nullptr }.

struct AFX_EXTENSION_MODULE
{
BOOL bInitialized;
HMODULE hModule;
HMODULE hResource;
CRuntimeClass* pFirstSharedClass;
COleObjectFactory* pFirstSharedFactory;
};


corob-msft (approved):

corob-msft:
LGTM. Thanks!

PR #2618, opened 2020-11-22
Three syntaxes for initialising auto-typed variables

yecril71pl:
The initialisation expression can be an assignment but that means `a = b = 0`.
The initialisation expression can involve `operator new` but it is not a separate case. Compare `auto a = new int;` and `auto a (new int);`.  There is no initialisation statement inherently involving `operator new` (unlike in Visual Basic).
The fact that a `for` loop involves an initialisation statement is a side note at best.  This trivia fact is not specific to `auto`.
The parenthesised "explanations" explained one unknown with another unknown.  I replaced them with short examples.

### DIFF
--- a/docs/cpp/auto-cpp.md
+++ b/docs/cpp/auto-cpp.md
@@ -40,7 +40,7 @@ Conversion cases in which you might not want to use **`auto`**:
 
 To use the **`auto`** keyword, use it instead of a type to declare a variable, and specify an initialization expression. In addition, you can modify the **`auto`** keyword by using specifiers and declarators such as **`const`**, **`volatile`**, pointer (**`*`**), reference (**`&`**), and rvalue reference (**`&&`**). The compiler evaluates the initialization expression and then uses that information to deduce the type of the variable.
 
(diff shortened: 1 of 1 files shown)


PRMerger13:

corob-msft:
@yecril71pl 
Thanks for your considered input. I'm not sure why `new` was called out either. Your updated explanation is correct as far as it goes, but it's missing a mention of `auto` and the uniform initialization syntax. We do mention it later, but it probably should be included here, too. It's safe to defer talk about how `auto` will create a `std::initializer_list` from a braced initializer at any opportunity. And there's a reason the range-based `for` loop syntax gets a call-out, since it's an `auto` syntax variation that doesn't appear elsewhere. I will think about this, but not for a while, since I'm on vacation for a week.

PRMerger8:

yecril71pl:
> And there's a reason the range-based `for` loop syntax gets a call-out, since it's an `auto` syntax variation that doesn't appear elsewhere.

I formally disagree but I have restored it since it is a typical use case.  You may have to adapt the terminology though.

PRMerger20:

corob-msft (approved):
@yecril71pl 
Thanks for your patience. I appreciate your contribution, even though I've substantially rewritten it to make our grammar checker happy. I'll merge this version. You're welcome to submit additional changes as PRs if you think more clarity is needed.

PR #2670, opened 2020-12-18
Update cpp-amp-overview.md

ShootingKing-AM:
Small Correction.


PRMerger7:
@ShootingKing-AM : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@ShootingKing-AM 
When I run this code, I get a result of 8, not an 11. That's what I expect, too, if elements are numbered (0,0,0), (0,0,1), (0,0,2), (0,0,3), (0,1,0), (0,1,1), (0,1,2), (0,1,3), etc. Is your result different?

ShootingKing-AM:
Oh Yes, true, my bad. I was mixing this code with something else.

PR #2781, opened 2021-02-16
Update abstract-classes-cpp.md

tiagomacarios:
update example to use a defaulted destructor

### DIFF
--- a/docs/cpp/abstract-classes-cpp.md
+++ b/docs/cpp/abstract-classes-cpp.md
@@ -54,11 +54,8 @@ This helps when designing class hierarchies whose base class(es) include pure vi
 class base {
 public:
     base() {}
-    virtual ~base()=0;
+    virtual ~base() = default;
 };
-
-// Provide a definition for destructor.
-base::~base() {}
 
 class derived:public base {
 public:


PRMerger16:
@tiagomacarios : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@tiagomacarios 
This change doesn't fit the surrounding description. A defaulted destructor means `base` is no longer an abstract class, and the desctructor is no longer pure virtual. What's the intent of your change?

tiagomacarios:
One of the engineers got confused by the example. I am as well to be honest. Indeed it says "pure virtual" but it goes and defines a destructor nevertheless:
```
// Provide a definition for destructor.
base::~base() {}
```
So why not modify the code to use the modern format?
Do you want me to also remove the "pure" mentions in the example?

corob-msft:
@tiagomacarios

Yes, the sample does define the `base` destructor even though it's been declared by using `= 0` in the class declaration, That's because if you don't implement it, you get a linker error in `~derived`, as the compiler generates code to invoke the destructors of the derivation tree in order from derived to base. The example illustrates this quirk of C++, and it's also pretty much the simplest possible abstract base class. 

The example doesn't have to use the `base` destructor to make the class abstract. It's possible to declare another member function as unimplemented pure virtual, and not involve the destructor. That wouldn't illustrate this weird destructor case, though.

corob-msft:
@tiagomacarios 
I'm closing this PR unmerged, since I don't think the change captures the intent of the document. However, don't think we don't value your contribution. I wouldn't have looked at the Abstract Classes topic without your input. I've rewritten it in a separate PR to try to make the example clearly about when and why you'd create *defined* pure virtual functions, not defaulted or deleted ones. It should show up tomorrow by noon Redmond time. Thanks for helping make the docs better.

PR #2989, opened 2021-02-25
Fix std::fputws return value documentation

Radfordhound:
The current Microsoft documentation for std::fputws incorrectly states that it "returns **WEOF**", when, in reality, it returns **EOF** as-per the C/C++ standard.

I've verified this by checking ucrt's "stdio\fputws.cpp", which returns **EOF** and also makes a note of this in a comment:
```
(Note well that we return EOF and not WEOF on failure.  This is intentional, and is the correct behavior per the C Language Standard).
```

### DIFF
--- a/docs/c-runtime-library/reference/fputs-fputws.md
+++ b/docs/c-runtime-library/reference/fputs-fputws.md
@@ -37,7 +37,7 @@ Pointer to **FILE** structure.
 
 ## Return Value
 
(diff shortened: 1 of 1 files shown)


PRMerger8:
@Radfordhound : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TylerMSFT (approved):
Thank you for improving the docs!

PR #2993, opened 2021-02-26
string -> stringBuffer on strcpy_s... page

cartwrightluke:
When this page is viewed on the microsoft documentation page (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s?view=msvc-160), the variable "string" is highlighted in code as if it were a type, and an array of type "wchar_t string" were being allocated. This is confusing - the "string" variable should be highlighted as if it were a variable.

I've renamed the "string" variable to "stringBuffer" to ensure it highlights correctly without changing the meaning of the code. However, I'm not familiar with this repository or the coding conventions used.

image

### DIFF
--- a/docs/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s.md
+++ b/docs/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s.md
@@ -143,19 +143,19 @@ Unlike production quality code, this sample calls the secure string functions wi
 
 int main(void)
 {
-    char string[80];
+    char stringBuffer[80];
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@cartwrightluke : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TylerMSFT (approved):
@cartwrightluke, thank you for improving the docs. I've approved this PR. 
We are having some trouble with our publishing pipeline at the moment, so it may be into next week before your changes show up in the live tree.

PR #2994, opened 2021-02-26
wstrlen -> wcslen on strlen... page

cartwrightluke:
The comment states that "wstrlen gives the length of a wide string", but then uses the "wcslen" function. The rest of the page indicates that "wcslen" is the correct function, so I modified the comment to match.


PRMerger19:
@cartwrightluke : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TylerMSFT (approved):
Thank you! 
Our content publishing pipeline is having some issues at the moment, but I hope you'll see your change in the live docs next week or so.

PR #3037, opened 2021-03-15
Fix dumpbin command switch

sheila-stewart:
The `/symbols` switch only outputs a summary; you need the `/exports` switch to list the decorated names.

### DIFF
--- a/docs/build/reference/decorated-names.md
+++ b/docs/build/reference/decorated-names.md
@@ -70,7 +70,7 @@ You can get the decorated form of a symbol name after you compile the source fil
 
 #### To use DUMPBIN to view decorated names
 
-1. To see the exported symbols in an .obj or .lib file, enter `dumpbin /symbols` `objfile` at a developer command prompt.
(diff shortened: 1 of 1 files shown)


PRMerger10:
@sheila-stewart : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:

corob-msft (approved):
@sheila-stewart Thanks for the update. The symbols might have this information, in a blizzard of other noisy reporting, but the exports usually has what you're looking for without the noise.

sheila-stewart:
@corob-msft The real reason I reported this was because I have a silly habit of running the dumpbin command against the DLL, instead of the LIB file. When run against a DLL, the `/symbols` switch doesn't list the exports, just a simple summary. It's a PEBCAK problem, but I've done this more than once!

Anyway, the added benefit of using `/exports` instead of `/symbols` is that you can get the exported symbols from the DLL _or_ a LIB.

PR #3041, opened 2021-03-17
Update C26432 doc - use default keyword everywhere

Thieum:
The "no warning" example doesn't use the default keyword everywhere it could be.

Other analyzers could complain: https://rules.sonarsource.com/cpp/RSPEC-3490

### DIFF
--- a/docs/code-quality/c26432.md
+++ b/docs/code-quality/c26432.md
@@ -30,7 +30,7 @@ Special operations such as constructors are assumed to alter the behavior of typ
 
 ## Example
 
-In this example, `warning::S` defines only a default constructor and a destructor. The `no_warning::S` declaration defines all five special member functions.
+In this example, `warning::S` defines only a default constructor and a destructor. The `no_warning::S` declaration defines or deletes all five special member functions.
 
 ```cpp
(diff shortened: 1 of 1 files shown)


PRMerger19:
@Thieum : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@Thieum 
Thanks for the contribution. You raise a good point about mixing user-defined but default-like special member functions and defaulted ones as raising a code smell. However, the change you suggest eliminates one code smell for another one; if all special member functions are explicitly defaulted, then it's better to not define any of them, and let them all be implicitly defaulted.

The warning description specifically says it's okay to mix definition types, so that's why the sample was written as it was. To make the example about adding the missing definitions rather than changing the existing ones, I've taken the liberty of modifying it slightly to do some explicit resource management. If that's okay with you, add your CLA signature, and I'll merge it.

corob-msft (approved):

PRMerger13:

Thieum:
@corob-msft Thanks a lot for your explanation! It does make sense, and it's really nice of you to take the time to improve my contribution!

PR #3054, opened 2021-03-25
Update walkthrough-compiling-a-native-cpp-program-on-the-command-line.md

laurenprinn:
I was a bit confused when I was first following along the tutorial because I couldn't find a way to save a Notepad file as a cpp extension, because Save As doesn't explicitly give that option. Hopefully this prevents others from having this confusion


PRMerger20:
@laurenprinn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger14:

corob-msft (approved):
@laurenprinn 
Thanks for the update. We're always on the lookout for ways our walkthroughs can fail. I hope you don't mind my expansion. Hopefully it won't trigger the CLA check.

PR #3068, opened 2021-04-03
Fix ComputeXdataSize samples for ARM/ARM64

AntonLapounov:
The ARM version was logically correct; however, its text was badly broken.  The ARM64 version had a number of logical issues:
- The `Xdata` parameter was typed incorrectly (the code would not compile).
- An incorrect shift amount (`27` instead of `22`) was used to check whether the extension word is present.  That contradicted the textual description.
- An incorrect mask (`0x0f` instead of `0x1f`) was used to extract the 5-bit `UnwindWords` field.  That caused losing the most significant bit of `UnwindWords` and returning an incorrect size for .xdata records containing from 61 to 124 unwind byte codes.
- Checking the `E` bit was missing.

See the implementation of the `RtlpUnwindFunctionFull` OS function for the source of truth.

### DIFF
--- a/docs/build/arm-exception-handling.md
+++ b/docs/build/arm-exception-handling.md
@@ -188,29 +188,32 @@ When the packed unwind format is insufficient to describe the unwinding of a fun
 The `.xdata` record is designed so that it is possible to fetch the first 8 bytes and compute the full size of the record, not including the length of the variable-sized exception data that follows. This code snippet computes the record size:
 
 ```cpp
-ULONG Comput`.xdata`Size(PULONG `.xdata`)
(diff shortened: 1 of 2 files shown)


PRMerger18:
@AntonLapounov : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft (approved):
@AntonLapounov The code in ARM exception handling looked like the victim of a global search and replace gone badly wrong. Thanks for the fixes.

AntonLapounov:
Thank you for the super-fast approval!

PR #3096, opened 2021-04-20
Update exception-handling-in-the-concurrency-runtime_3.cpp

sapant-msft:
typo fixed


corob-msft (approved):
@sapant-msft 
Good catch, thanks for the update.

PR #3128, opened 2021-05-10
extraneous named parameters on deleted functions

Thieum:
Unless it's part of a coding standard, I think named parameters can be skipped for deleted functions.


corob-msft (approved):
@Thieum Looks reasonable. Thanks for the update.

PR #3130, opened 2021-05-11
It's showed as C++ on docs.microsoft.com and is confusing.

0xbadfca11:


corob-msft (approved):
@0xbadfca11 Looks good, thanks for the update.

PR #3190, opened 2021-06-10
correct "friend class" syntax

mnatsuhara:
The syntax `friend class F` is described toward the top of the page, but in these two spots the `friend` and `class` are I believe incorrectly transposed.  `class friend F` is incorrect syntax.


PRMerger9:
@mnatsuhara : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3193, opened 2021-06-12
Removing reference to /Oa

manbearian:
The `/Oa` switch hasn't been in the compiler for more than a decade (nearly two decades). This reference to /Oa is, i believe, the only reference to this long gone switch; i believe referencing /Oa in this context is actually more harmful then helpful.


PRMerger20:
@manbearian : Thanks for your contribution! The author(s) have been notified to review your proposed change.

GabLotus:
image
Validated the this is true

corob-msft (approved):
Seems reasonable.

corob-msft:
Looks like /Oa was removed in 2005. It was silently ignored for at least a while according to [Visual C++ What's New 2003 through 2015](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015) but at some point it must have been dropped, so it generates a D9002 now.

corob-msft:
#sign-off

PR #3211, opened 2021-06-19
Update Walkthrough Creating traditional Windows Desktop Apps.

thispsj:
This PR updates [this](https://docs.microsoft.com/cpp/windows/walkthrough-creating-windows-desktop-applications-cpp?view=msvc-160) article on ms docs. The code has been updated to use `CreateWindowEx`.


PRMerger17:
@thispsj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@thispsj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

thispsj:
@corob-msft any problems with the PR ?

corob-msft:
@thispsj 
Thanks for bringing this sample from the late '80s into the early '90s. I've been vacationing and not paying close attention to the repo. Your changes are mostly fine, though I don't see much value in changing the `wcex.hIcon` load from using the parameter `hInstance` to using the copy in the struct. In optimized code, the value is enregistered at that point, so it makes no difference, while in debug code it could conceivably cause an extra memory access. (I haven't checked the assembly output to verify.) Did you have a purpose or style guideline in mind for that change?

thispsj:
@corob-msft Here are the assembly output for both cases in debug mode.
Screenshot (62)
Screenshot (63)

This is one of my small projects which uses `WCNDCLASSEX` structure to create the window. Here the variable assigned to the structure is `wcx` and not `wcex`. That's the only difference in the cpp code.
As you see in the assembly, the `mov` instruction directly registers the pointer to `hInst` (the `HINSTANCE` variable) when directly passed while when `wcex.hInstance` is used it passes that pointer. Otherwise it's the same.

And one more thing, the sample code is itself using `wcex.hInstance` to load the small icon. So, to my main purpose was to use the same logic as applicable everywhere because I've seen people asking that why is it different in the same code (i.e. at one place `hInst` and at the another `wcex.hInstance`.

corob-msft (approved):
Looks good, @thispsj - thanks for the update.

PR #3219, opened 2021-06-23
Update cmake-linux-configure.md

FrankAtHexagon:
Added information about change from Unix Makefiles so that example JSON is less confusing.

### DIFF
--- a/docs/linux/cmake-linux-configure.md
+++ b/docs/linux/cmake-linux-configure.md
@@ -85,7 +85,7 @@ The default Linux-Debug configuration in Visual Studio 2019 version 16.1, and la
 }
 ```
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@FrankAtHexagon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TylerMSFT:
@FrankAtHexagon , thank you for taking an interest in updating the docs. I'm interested in your thinking behind adding "as opposed to Unix Makefiles". I'm curious why you think that emphasis needs to be added. The text right now says what you need, without introducing a secondary concept that may make some readers wonder why it is being called out.

FrankAtHexagon:
The reason I thought this was necessary is because the text above the example says this:
> The default Linux-Debug configuration in Visual Studio 2019 version 16.1, and later, looks like this...

And then provides an example with 'Unix Makefiles.' However, the text right after the example says this:
> In Visual Studio 2019 version 16.6 or later, Ninja is the default generator for configurations targeting a remote system or WSL.

So I thought it was confusing which generator would typically be the default, until I read the blog post.

TylerMSFT:
@FrankAtHexagon , great justification. Thank you for explaining. I'll get your change merged today. I really appreciate you taking the time to improve our docs.

TylerMSFT (approved):

PR #3223, opened 2021-06-27
Update template-specialization-cpp.md

Jaiganeshkumaran:
Adds the delete statement for showcasing good practice.

### DIFF
--- a/docs/cpp/template-specialization-cpp.md
+++ b/docs/cpp/template-specialization-cpp.md
@@ -285,25 +285,23 @@ public:
 };
 
 int main() {
-   auto dict = new Dictionary<char*, char*>(10);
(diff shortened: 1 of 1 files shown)


PRMerger13:
@Jaiganeshkumaran : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger15:

corob-msft (approved):
@Jaiganeshkumaran Thanks for the update and for bringing the leaky code to my attention. Rather than use heap objects, I edited your changes to use Modern stack-based style, so the objects are deleted when they go out of scope. I also updated the template parameters to match the current C++ string literal definitions. This code would have remained broken without your submission.

PR #3228, opened 2021-06-29
Typo?

Brian-Taylor8:
I think here it's meant to say that int is for integer values, not integral values! Really useful docs by the way!


PRMerger7:
@Brian-Taylor8 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@Brian-Taylor8 You're right, "integer values" is more appropriate than "integral values" here, because it's more likely to be translated correctly. It wouldn't do for the translated version to be turned into something like "`int` is for essential values." If you want to reopen this, I'll approve it.

Brian-Taylor8:
Thanks! I deleted the fork, so I'll resubmit!

PR #3237, opened 2021-07-02
Making /Qspectre 'Required Libraries' section more clear

ericbrumer:
A C++ customer opened a DevCom ticket (https://developercommunity.visualstudio.com/t/Build-unexpectedly-succeeds-with-Qspect/1437998) indicating some confusion with what happens when the `/Qspectre` switch is thrown. The compiler switch only modifies the compiler, and our MSBuild properties for C++ projects is what links in the additional libraries.

This PR attempts to clear that up.

### DIFF
--- a/docs/build/reference/qspectre.md
+++ b/docs/build/reference/qspectre.md
@@ -52,21 +52,23 @@ The effect of **`/Qspectre`** on performance appeared to be negligible in severa
 
 ### Required libraries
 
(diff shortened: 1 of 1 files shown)


corob-msft (approved):

corob-msft:
I may need to tweak something when this gets to the publishing system if the build checks find a problem, but it's good enough to check in here. Thanks @ericbrumer for the contribution!

PR #3249, opened 2021-07-12
UD2 is allowed in Ring 3 code

AzAgarampur:
Sometimes I'll use __ud2() as a way of intentionally triggering an exception in order to execute a SEH exception handler, rather than triggering one with something like `*(int*)0;`.


PRMerger19:
@AzAgarampur : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft (approved):
@AzAgarampur Thanks for the update.

PR #3276, opened 2021-08-02
Updated C/C++ workload name

ainrd:
The workload for C/C++ development in VS Build Tools 2019 installer is called `Desktop development with C++`.

https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019#desktop-development-with-c

The old name is probably from the previous release. 

A small update that might help avoid confusion.


PRMerger20:
@ainrd : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft (approved):
@ainrd Thanks for the update. It looks like names have changed in both the current and previous versions, so I called that out.

PR #3278, opened 2021-08-03
Fix the variable name in Example 2

ScottHutchinson:
There is no sp1 in the example. It should be sp7 instead.


corob-msft:
@ScottHutchinson 
Thanks for taking the time to propose a change for the docs. Looking closely at it, I don't see the need for this change. There's no `sp2` defined in the second example either, if you're only considering the local context. The article explains how the snippets go together. The snippet in Example 1 declares and defines both `sp1` and `sp2`, which then get used in Example 2.  There's nothing wrong with your updated code, but it seems to be fixing a problem that isn't really a problem. In the interest of not fixing what isn't broken, I'm closing the PR. However, please don't let this discourage you from proposing other fixes, or if you think I'm wrong, for arguing your case here.

PR #3284, opened 2021-08-07
Setting PATH allows DUMPBIN to be executed from the command prompt

d-c-d:
There are multiple DUMPBIN.EXE's in the MSVC folder, but they can be executed if the PATH is set correctly.
If this is not true, or undesireable per system design constraints, please clarify otherwise.


PRMerger14:
@d-c-d : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger14:

corob-msft (approved):
@d-c-d 
Thanks for the contribution. It's true, the blanket statement was overbroad. I've expanded your addition to provide a link to instructions for adding the required environment values to a command prompt.

PR #3353, opened 2021-09-06
Fix punctuation

veganaize:


PRMerger19:
@veganaize : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@veganaize 
Thanks for taking the time to submit this pull request. The best submissions for fixes include a description of the problem and a justification for the change (often the same thing). There's more than a simple punctuation fix going on here. It looks like the original tried to create a list with partial sentences. A better fix would use a colon and gather up the list elements using commas or semicolons. Since I have more leeway in determining what we want to convey, I'd express it now with this sentence:

> C++ can create almost any kind of program: Games, device drivers, HPC, cloud, desktop, embedded, and mobile apps, and much more.

Does this satisfy the issue you wanted to address?

veganaize:
No.

PR #3356, opened 2021-09-08
updated mfc heirarchy chart 2 of 3

AnjuDel:
In chart 2 of 3 the “PropertyPages” were hanging from nowhere. They should hang under the “CDialog”


AnjuDel:
mfc_hierarchy_chart2of3
This is how the new file looks

corob-msft:
@AnjuDel 
Thanks for the update! Let's move this PR to the private repo. I'd like to rename the file and change the links that reference it. I'll send email to follow up.

corob-msft:
@AnjuDel 
I'm closing this PR in favor of https://github.com/MicrosoftDocs/cpp-docs-pr/pull/3769 instead. If you don't have access to cpp-docs-pr, see the [Docs Contributor Guide here](https://review.docs.microsoft.com/en-us/help/contribute/contribute-get-started-setup-github?branch=main).

AnjuDel:
Hi,

Thanks for closing this and completing the pr. Sorry for the late response I was already out on Friday. I didn't know we have a public and a private repo. What's the difference between these 2?

Thanks,
Anju

corob-msft:
@AnjuDel 
We have a private repo, cpp-docs-pr, that we use for staging and publishing to the live site. It runs extra validation checks on the content and generates an internally visible high-fidelity preview for proofing, so we prefer using it for non-trivial updates. It's only accessible to members of the Microsoft group (employees and vendors). We sync it with the public site after every daily release and after every public contribution, so everyone sees the same content and can make the same changes in either repo. It's just easier to verify the content isn't broken by a change when it's in the private repo and we can see it on the review site.

PR #3403, opened 2021-09-24
Update understanding-sal.md

mohammad-ghasemi-2549:
I found these slides from a lecture from a Microsoft employee about SAL very helpful for both visualizing and recalling the SAL concepts.
For newcomers it is hard to remember what each SAL annotation means. There are visualizations of the meaning of important annotations in these slides.

### DIFF
--- a/docs/code-quality/understanding-sal.md
+++ b/docs/code-quality/understanding-sal.md
@@ -394,6 +394,8 @@ Here are some guidelines:
 
 Or you can annotate all parameters to make your intent clear throughout and to make it easy to check that annotations have been done.
 
+Diagrams in [these slides](https://www.cs.cmu.edu/~fp/courses/15122-f10/lectures/25-sal.pdf) are very helpful for visualizing and recaling SAL concepts.
+
 ## See also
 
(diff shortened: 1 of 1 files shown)


PRMerger16:
@mohammad-ghasemi-2549 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

corob-msft:
@mohammad-ghasemi-2549 

This is a nice set of lecture notes you've found. The content is useful and informative. And unfortunately, I'm not going to add it.

The issue I have is, this link isn't the most useful way to share info with readers of our content. It would be better to make sure our own content includes these descriptions and insights where they make the most sense. Then readers don't have to go through 50+ PowerPoint slides in an obscure set of lecture notes that might go away without warning to get that info. As you might imagine, this is a much bigger task than just adding a link to one of our pages. It would be nice if we had someone with the time to take that task on.

As a general rule, we try to avoid external links that are likely to become maintenance problems in the future. A deep link into college course materials seems very likely to fit that description, even if it's been around for a decade. So, I appreciate your contribution, and I like what the material has to say, but I don't think this is the way we should present it to our readers.

PR #3447, opened 2021-10-10
Update cmake-presets-vs.md

### LINKED ISSUE #3446: Update CMake Presets for compiling with clang-cl
Currently the [instructions](https://docs.microsoft.com/en-us/cpp/build/cmake-presets-vs?view=msvc-160) for configuring clang-cl builds in Visual Studio with CMake Presets are as follows:

```json
"cacheVariables": {
  "CMAKE_BUILD_TYPE": "Debug",
  "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
  "CMAKE_C_COMPILER": "clang-cl",
  "CMAKE_CXX_COMPILER": "clang-cl"
},
```

However I'm wondering why the `toolset` configure preset is not used instead like so:

```json
"cacheVariables": {
  "CMAKE_BUILD_TYPE": "Debug",
  "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}"
},
"toolset": "ClangCL",
```

As far as I understand, this approach leverages the schema of CMake Presets more directly, so should be preferred.

Also this approach has the added benefit of working in Visual Studio Code as well with no additional configuration required. (The current instructions to use `CMAKE_CXX_COMPILER` do not work in Visual Studio Code or on the command line without `vcvarsall.bat`).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 18671a37-b8d9-b8cd-51fc-c1cd267e6af5
* Version Independent ID: 6f0932ad-998a-9449-dfff-259452ba20d5
* Content: [Configure and build with CMake Presets](https://docs.microsoft.com/en-us/cpp/build/cmake-presets-vs?view=msvc-160)
* Content Source: [docs/build/cmake-presets-vs.md](https://github.com/Microsoft/cpp-docs/blob/master/docs/build/cmake-presets-vs.md)
* Product: **visual-cpp**
* Technology: **cpp-tools**
* GitHub Login: @corob-msft
* Microsoft Alias: **corob**

sweemer:
Add instructions for configuring clang-cl with the ClangCL toolset configure preset instead of by setting CMAKE_C_COMPILER and CMAKE_CXX_COMPILER in cacheVariables.

Fixes #3446.

### DIFF
--- a/docs/build/cmake-presets-vs.md
+++ b/docs/build/cmake-presets-vs.md
@@ -223,7 +223,7 @@ Build with `clang`:
 }
 ```
 
-*OR* use the `toolset` configure preset to specify the `ClangCL` toolset like so:
+*OR* if you are using a generator that supports specifying a native build system toolset, such as Visual Studio Generators for VS 2010 and above, then you can use the `toolset` configure preset to specify the `ClangCL` toolset like so:
   
 ```json
 "cacheVariables": {
@@ -238,7 +238,9 @@ Build with `clang`:
   }
 }
 ```
-  
+
+See the documentation on [CMAKE_GENERATOR_TOOLSET](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html) for more information on generators that support the `toolset` specification.
+
 > [!IMPORTANT]
 > In Visual Studio 2019, you must explicitly specify a Clang IntelliSense mode when you're building with `clang` or `clang-cl`.
 
--- a/docs/build/cmake-presets-vs.md
+++ b/docs/build/cmake-presets-vs.md
@@ -223,8 +223,8 @@ Build with `clang`:
 }
 ```
 
-*OR* if you are using a generator that supports specifying a native build system toolset, such as Visual Studio Generators for VS 2010 and above, then you can use the `toolset` configure preset to specify the `ClangCL` toolset like so:
-  
+*OR* if you are using either `Visual Studio 16 2019` or `Visual Studio 17 2022` as your generator, then you can use the `toolset` configure preset to specify the `ClangCL` toolset:
+
 ```json
 "cacheVariables": {
   "CMAKE_BUILD_TYPE": "Debug",
--- a/docs/build/cmake-presets-vs.md
+++ b/docs/build/cmake-presets-vs.md
@@ -223,6 +223,22 @@ Build with `clang`:
 }
 ```
 
+*OR* use the `toolset` configure preset to specify the `ClangCL` toolset like so:
+  
+```json
+"cacheVariables": {
+  "CMAKE_BUILD_TYPE": "Debug",
+  "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
+},
+"toolset": "ClangCL",
+
+"vendor": {
+  "microsoft.com/VisualStudioSettings/CMake/1.0": {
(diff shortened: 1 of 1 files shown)


PRMerger15:
@sweemer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@bobbrow Do you have a delegate who can review this change and the related issue #3446? I've let this go too long without a ping.

sfreed141:
I commented on the original issue too, but I don't believe ClangCL is a valid toolset for the Visual Studio generators.
image

@sweemer did using ClangCL as the toolset work for you? If so can you share your VS and CMake versions?

docs/build/cmake-presets-vs.md
```diff
@@ -223,6 +223,22 @@ Build with `clang`:
...
 ```
 
+*OR* use the `toolset` configure preset to specify the `ClangCL` toolset like so:
+  
+```json
+"cacheVariables": {
+  "CMAKE_BUILD_TYPE": "Debug",
+  "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
+},
+"toolset": "ClangCL",
```
sfreed141, line 233: This assumes the current `generator` supports specifying toolsets (see [docs](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html) for more info). For the `Ninja` generator this preset is invalid.

PRMerger12:
@sweemer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/build/cmake-presets-vs.md
sweemer: You're right. I have pushed up a new commit to clarify that not all generators support this preset.

sfreed141:
Just copying this over here from the issue:
I dug into why it wasn't working for me and it's because of microsoft/vcpkg#15623 :( After I removed vcpkg using ClangCl for the toolset works. I believe it only works for the VS generators though (based on [this bit in CMake source](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/CMakeDetermineCompilerId.cmake#L340))

docs/build/cmake-presets-vs.md
sfreed141: Thanks for updating that. However I don't think ClangCl will be valid until the `Visual Studio 16 2019` generator, since LLVM was added in [version 16.1](https://devblogs.microsoft.com/cppblog/clang-llvm-support-in-visual-studio/). Also I'd be more explicit that this is only for Visual Studio generators. perhaps

> if you are using either `Visual Studio 16 2019` or `Visual Studio 17 2022` as your generator, you can use the `toolset` property in the configure preset to specify the `ClangCL` toolset:

PRMerger6:
@sweemer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/build/cmake-presets-vs.md
sweemer: Done

sfreed141 (approved):

PRMerger4:

corob-msft (approved):

corob-msft:
@sweemer Thanks for the contribution, and for your patience and collaboration. Thanks to @sfreed141 too, for verifying and suggestions.

PR #3458, opened 2021-10-15
Update download URLs to point to 2022 redists

dxgldotorg:
The docs say 2022 however it was pointing to the aka.ms link for version 16 not 17.  Since all three version 17 downloads are live as of the RC (which is go-live licensed), I edited this file to reflect the changes.


PRMerger12:
@dxgldotorg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

dxgldotorg:
Just in the process of migrating from 2019 to 2022 RC as it is now go-live, and saw the outdated document.

corob-msft (approved):
@dxgldotorg Thanks for the updates.

PR #3475, opened 2021-10-26
Add name clarification to new description

wmcnamara:
This PR adds a line that refers to the free store as the the "heap", to clarify for people who may have only heard the term "heap", and not "free store".


corob-msft (approved):
@wmcnamara 
Thanks for the update. It's correct as far as it goes, though `new` in placement form can also target the stack. I'm guessing the "free store" term comes from the original draft of what eventually became ISO C++, the Annotated 
C++ Reference Manual, which was relied on heavily for the original C++ language docs on MSDN, some of which are largely unchanged since 1992.

PR #3509, opened 2021-11-09
Remove the deprecation of IT instructions in the ARM ABI

mstorsjo:
Upstream ARM has undeprecated the use of IT instructions. There's no formal announcement of the matter in itself, but the previous deprecations are being removed from the official docs.

See https://reviews.llvm.org/D108679#2975603 for a discussion on the matter, CC @dpaoliello who was involved in that discussion.

### DIFF
--- a/docs/build/overview-of-arm-abi-conventions.md
+++ b/docs/build/overview-of-arm-abi-conventions.md
@@ -38,28 +38,6 @@ The instruction set for Windows on ARM is strictly limited to Thumb-2. All code 
 
 A side-effect of this requirement is that all code pointers must have the low bit set. Then, when they're loaded and branched to via BLX or BX, the processor remains in Thumb mode. It doesn't try to execute the target code as 32-bit ARM instructions.
 
-### IT Instructions
-
-The use of IT instructions in Thumb-2 code is disallowed except for these specific cases:
-
-- The IT instruction can only be used to modify one target instruction.
-
-- The target must be one of these 16-bit instructions:
-
-   |16-Bit Opcodes|Class|Restrictions|
-   |---------------------|-----------|------------------|
-   |MOV, MVN|Move|`Rm != PC`, `Rd != PC`|
-   |LDR, LDR[S]B, LDR[S]H|Load from memory|But not LDR literal forms|
-   |STR, STRB, STRH|Store to memory||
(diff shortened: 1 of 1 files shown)


PRMerger16:
@mstorsjo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

corob-msft:
@mstorsjo and cc: @dpaoliello 
Do you know if this un-deprecation (is that a word?) applies to previous versions of our armasm tool, and to all ARM processors? Just wondering if there's a set of processors or old assemblers where the deprecation still holds, especially if certain code won't work on Windows on ARM.

dpaoliello:
> Do you know if this un-deprecation (is that a word?) applies to previous versions of our armasm tool, and to all ARM processors? Just wondering if there's a set of processors or old assemblers where the deprecation still holds, especially if certain code won't work on Windows on ARM.

@corob-msft The un-deprecation is a change by ARM - there have been no changes that I know of to any Microsoft tools or OSes.

From the Windows side, these instructions were always supported, and I don't believe that we ever supported any hardware that didn't support the instructions.

mstorsjo:
> @mstorsjo and cc: @dpaoliello
> Do you know if this un-deprecation (is that a word?) applies to previous versions of our armasm tool, and to all ARM processors? Just wondering if there's a set of processors or old assemblers where the deprecation still holds, especially if certain code won't work on Windows on ARM.

When it comes to actual processors, there hasn't been any issue so far (as it would break backwards binary compatbility), but I've understood it that they (ARM) reserved the right to degrade/break it later (and thus put in the deprecation preemptively) - but now they've decided that they won't do that, so there's no longer any reason to avoid those constructs.

Yeah ideally armasm would also stop warning about these constructs - I guess I could file a public bug report about it, but I'm not sure if it's as effective as if someone of you would start it.

> From the Windows side, these instructions were always supported, and I don't believe that we ever supported any hardware that didn't support the instructions.

Yeah I don't think there has been any (armv8-a) processors that didn't support those instructions (and the deprecation was added as part of armv8 to begin with).

corob-msft (approved):
@mstorsjo Thanks for the update and the info. Merging shortly. Docs should be updated tomorrow.

PR #3526, opened 2021-11-17
Correct the documentation for fmaf

vtjnash:
The documentation states that this function is correctly rounded, but empirically, that is not true, so here we remove that incorrect information from the documentation. For example, this would be the correct answer, but it is not the answer that is given by this function (unless the computer has FMA3 hardware).

fmaf(-1.9369631e13f, 2.1513551e-7f, -1.7354427e-24f) == -4.16709575e6

Some sample test code to demonstrate why this documentation is wrong:
```c
#include <cmath>
#include <iostream>
int main() {
    float a = -1.9369631e13f, b = 2.1513551e-7f, c = -1.7354427e-24f;
    volatile float va = a, vb = b, vc = c;
    float f1 = fmaf(a, b, c);
    float f2 = fmaf(va, vb, vc);
    std::cout.setf(std::ios::scientific, ::std::ios::floatfield);
    std::cout.precision(19);
    std::cout << f1 << " = -4.1670957500000000000e+06\n";
    std::cout << f2 << " = -4.1670957500000000000e+06\n";
    return 0;
}
```

This outputs:
```
-4.1670955000000000000e+06 = -4.1670957500000000000e+06
-4.1670955000000000000e+06 = -4.1670957500000000000e+06
```

Which we can see means the documentation is wrong, since that `5` needs to be a `7` for this function to be corrected rounded.

### DIFF
--- a/docs/c-runtime-library/reference/fma-fmaf-fmal.md
+++ b/docs/c-runtime-library/reference/fma-fmaf-fmal.md
@@ -1,6 +1,6 @@
 ---
 title: "fma, fmaf, fmal"
-description: "API reference for fma, fmaf, and fmal; which multiplies two values together, adds a third value, and then rounds the result, without losing any precision due to intermediary rounding."
+description: "API reference for fma, fmaf, and fmal; which multiplies two values together, adds a third value, and then rounds the result, while only losing a small amount of precision due to intermediary rounding."
 ms.date: "9/1/2020"
 api_name: ["fma", "fmaf", "fmal", "_o_fma", "_o_fmaf", "_o_fmal"]
 api_location: ["msvcrt.dll", "msvcr80.dll", "msvcr90.dll", "msvcr100.dll", "msvcr100_clr0400.dll", "msvcr110.dll", "msvcr110_clr0400.dll", "msvcr120.dll", "msvcr120_clr0400.dll", "ucrtbase.dll", "api-ms-win-crt-math-l1-1-0.dll", "api-ms-win-crt-private-l1-1-0.dll"]
@@ -12,7 +12,7 @@ ms.assetid: 584a6037-da1e-4e86-9f0c-97aae86de0c0
 ---
 # fma, fmaf, fmal
 
-Multiplies two values together, adds a third value, and then rounds the result, without losing any precision due to intermediary rounding.
+Multiplies two values together, adds a third value, and then rounds the result, while only losing a small amount of precision due to intermediary rounding.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


PRMerger8:
@vtjnash : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

ViralBShah:
This has been broken for years. It would be really nice if `fmaf` could be fixed to be correctly rounded.

TylerMSFT:
Hi, thank you for raising this issue. I'm talking this over with the dev team because it may not be as straightforward as the suggested fix implies. Will update this once we land on a resolution.

ViralBShah:
Thank you. FWIW, Julia has a [correct implementation](https://github.com/JuliaLang/julia/commit/ee36c1347ac5b313b843af7271ace049417038e3) now in its `fmaf_emulated` function along with test cases, if you'd like to use that as a reference.

oscardssmith:
Here are OpenLibm's versions (which are correct and support all rounding modes at the expense of some speed)
https://github.com/JuliaMath/openlibm/blob/master/src/s_fma.c
https://github.com/JuliaMath/openlibm/blob/master/src/s_fmaf.c

Here are the Glibc versions which are also correct and support rounding modes (warning GNU license)
https://github.com/lattera/glibc/blob/master/sysdeps/ieee754/dbl-64/s_fma.c
https://github.com/lattera/glibc/blob/master/sysdeps/ieee754/dbl-64/s_fmaf.c

ViralBShah:
I am not sure openlibm is fully correct: https://github.com/JuliaMath/openlibm/issues/160

ShannonLeavitt:
@TylerMSFT 

Can you review this old PR and determine whether it needs to be closed or merged?

oscardssmith:
This PR is still necessary to merge since Microsoft's docs still incorrectly claim that fma is correctly rounded despite it not being correctly rounded.

TylerMSFT:
I can take care of this today. I did get a bug opened on the CRT to address this problem.

But before we merge this, could update it to note that you lose precision unless the computer has FMA3 hardware?

oscardssmith:
A lot of the problem is that this isn't true. Even if your computer has FMA3 or FMA4 hardware, compilers will sometimes evaluate the wrong result at compile time.

TylerMSFT (approved):
There is a note in the bug I opened about this issue to come back and update this page once this is fixed.

TylerMSFT:
#sign-off

PR #20, opened 2017-04-07
More detail about __WRL_CONFIGURATION_LEGACY__

BenJKuhn:
This adds clarification to exactly what __WRL_CONFIGURATION_LEGACY__  changes.

PR #3541, opened 2021-11-21
Fix two incorrect conditions in the ARM packed exception handling

mstorsjo:
For the prologue, the distinction between `mov r11, sp` and
`add r11, sp, #xx` lies only in whether other registers are pushed
before R11 and LR in prologue instruction 2. (The condition
`R`==1 and `PF`==0 corresponds to no other registers pushed before
`r11`.) The `L` flag doesn't affect it (and the documentation explicit
says that when `C` is set, then `L` also must be set, thus `C`==1
and `L`==0 is tautological too).

This was correctly described in the textual form, but the conditions
in the table included one erronous case.

For the epilogue, when `Ret`!=0, the function doesn't return by
popping LR directly into PC, but with an explicit branch instruction
(10a or 10b). In this case, if `L`==1, the LR register is included
in the list of registers popped in instruction 8. If register
homing is used (`H`==1), then instruction 9a is used instead of 9b.

This case seemed to have been overlooked both in the textual
description and in the table.

CC @ThomsonTan who might be familiar with this area.

### DIFF
--- a/docs/build/arm-exception-handling.md
+++ b/docs/build/arm-exception-handling.md
@@ -81,7 +81,7 @@ Due to possible redundancies in the encodings above, these restrictions apply:
 
 - If the *`C`* flag is set to 1:
 
-  - The *`L`* flag must also be set to 1, because frame chaining required both r11 and LR.
+  - The *`L`* flag must also be set to 1, because frame chaining requires both r11 and LR.
 
   - r11 must not be included in the set of registers described by *`Reg`*. That is, if r4-r11 are pushed, *`Reg`* should only describe r4-r10, because the *`C`* flag implies r11.
 
@@ -101,8 +101,8 @@ Prologues for canonical functions may have up to 5 instructions (notice that 3a 
 |-----------------|-----------------------------------|----------|------------|------------------|
 |1|*`H`*==1|16|`push {r0-r3}`|04|
 |2|*`C`*==1 or *`L`*==1 or *`R`*==0 or *`PF`*==1|16/32|`push {registers}`|80-BF/D0-DF/EC-ED|
-|3a|*`C`*==1 and (*`L`*==0 and *`R`*==1 and *`PF`*==0)|16|`mov r11,sp`|C0-CF/FB|
-|3b|*`C`*==1 and (*`L`*==1 or *`R`*==0 or *`PF`*==1)|32|`add r11,sp,#xx`|FC|
+|3a|*`C`*==1 and (*`R`*==1 and *`PF`*==0)|16|`mov r11,sp`|FB|
+|3b|*`C`*==1 and (*`R`*==0 or *`PF`*==1)|32|`add r11,sp,#xx`|FC|
 |4|*`R`*==1 and *`Reg`* != 7|32|`vpush {d8-dE}`|E0-E7|
 |5|*`Stack Adjust`* != 0 and *`PF`*==0|16/32|`sub sp,sp,#xx`|00-7F/E8-EB|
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@mstorsjo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

corob-msft:
@mstorsjo 
Thanks, Martin. The updates look good to me, but since I have only the barest familiarity with ARM exception unwinding , I need to defer to experts on the development team to review. In addition to @ThomsonTan , @pmsjt, @VCZYK, or @RussKeldorph may know, or possibly know who knows.

ThomsonTan:
Thanks, Martin. I defer this to @pmsjt.

pmsjt:
Thanks for contributing. I'll look at this as soon as I can. It might take a couple of weeks tho.

I appreciate the patience.

PRMerger8:
@mstorsjo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
Ping @pmsjt for a review before we all go on vacation. :)

docs/build/arm-exception-handling.md
```diff
@@ -139,17 +139,17 @@ The epilogues for canonical functions follow a similar form, but in reverse and
 |-----------------|-----------------------------------|----------|------------|
 |6|*`Stack Adjust`*!=0 and *`EF`*==0|16/32|`add   sp,sp,#xx`|
 |7|*`R`*==1 and *`Reg`*!=7|32|`vpop  {d8-dE}`|
-|8|*`C`*==1 or (*`L`*==1 and *`H`*==0) or *`R`*==0 or *`EF`*==1|16/32|`pop   {registers}`|
-|9a|*`H`*==1 and *`L`*==0|16|`add   sp,sp,#0x10`|
-|9b|*`H`*==1 and *`L`*==1|32|`ldr   pc,[sp],#0x14`|
+|8|*`C`*==1 or (*`L`*==1 and (*`H`*==0 or *`Ret`* !=0)) or *`R`*==0 or *`EF`*==1|16/32|`pop   {registers}`|
+|9a|*`H`*==1 and (*`L`*==0 or *`Ret`*!=0)|16|`add   sp,sp,#0x10`|
+|9b|*`H`*==1 and *`L`*==1 and *`Ret`*==0|32|`ldr   pc,[sp],#0x14`|
```
pmsjt, line 144: In the Ret==0 case, return is made by pop {pc}, so it can't be the case here where return is being achieved by ldr pc, ...

pmsjt, line 144: Should we also add an explanation to the Ret==0 case saying that it could be pop {pc} OR it could also be ldr pc,[sp],#0x14?

pmsjt (changes requested):

pmsjt:
The truth table for the registers saved is also not great. Lines [127, 130] refer to C=1, L=0 which is an invalid configuration. I wonder if we should just replace the cells on those 4 lines with "Invalid Encoding" or something like that.

pmsjt:
Since you are already making changes here, I wonder if you would also mind fixing line 84: "required" -> "requires".

docs/build/arm-exception-handling.md
mstorsjo: On instruction 8? Maybe, but I'm not sure how to include that information without blowing up the table too much. Because if `R==0` then you'd still have instruction 8 popping a number of registers, but not including `pc`. The logical condition, while complex, should be accurate for when this instruction is present though.

mstorsjo:
> The truth table for the registers saved is also not great. Lines [127, 130] refer to C=1, L=0 which is an invalid configuration. I wonder if we should just replace the cells on those 4 lines with "Invalid Encoding" or something like that.

That would be good for consistency, yes, otherwise those lines can be misleading.

> Since you are already making changes here, I wonder if you would also mind fixing line 84: "required" -> "requires".

Sure.

docs/build/arm-exception-handling.md
```diff
@@ -139,17 +139,17 @@ The epilogues for canonical functions follow a similar form, but in reverse and
...
 
-If *`H`* is set, then either instruction 9a or 9b is present. Instruction 9a is used when *`L`* is 0, to indicate that the LR is not on the stack. In this case, the stack is manually adjusted and *`Ret`* must be 1 or 2 to specify an explicit return. Instruction 9b is used when *`L`* is 1, to indicate an early end to the epilogue, and to return and adjust the stack at the same time.
+If *`H`* is set, then either instruction 9a or 9b is present. Instruction 9a is used when *`Ret`* is nonzero, when instruction 10a or 10b is present (if *`L`*  is 1, it is popped as part of instruction 8). In this case, the stack is manually adjusted. Instruction 9b is used when *`L`* is 1 and *`Ret`*  is zero, to indicate an early end to the epilogue, and to return and adjust the stack at the same time.
```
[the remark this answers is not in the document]
mstorsjo: That sounds good, changing it that way.

PRMerger20:
@mstorsjo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mstorsjo:
I applied most of the suggested changes, except for the extra explanation about there either being `pop {pc}` or `ldr pc, [sp], #20` (as I'm unsure where you'd want that change and how it would fit in).

pmsjt (approved):
Thank you so much for your contribution. Can't believe how badly broken this was, with contradictions no more than a paragraph apart.

corob-msft (approved):

corob-msft:
Since this is a useful update as is, I'll leave the expansion on 9b for another update, and merge this now. Many thanks to @mstorsjo for your contributions and collaboration, and to @pmsjt for reviews and comments.

PR #3572, opened 2021-12-04
Changing confusing wording about C++ exceptions

hannes-harnisch:
The text mentions `std::throw`, but such a thing does not exist in the C++ standard library and cannot exist at all because `throw` is a reserved keyword.


PRMerger4:
@hannes-harnisch : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger4:

corob-msft (approved):
@hannes-harnisch Good catch. Thanks for taking the initiative to make an update.

PR #3658, opened 2022-01-19
Add note about `_CRT_DECLARE_NONSTDC_NAMES`

eldakesh-ms:
With `/c11` and `/c17` properly defining `__STDC__`, users may encounter that the POSIX functions they are relying on are no longer exposed. This was always the case with `/Za` but is now also visible with the standard C versions.

Gated by 17.2 preview 2 at the earliest.


PRMerger17:
@eldakesh-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger4:

corob-msft (approved):
@eldakesh-ms Looks like a useful addition. Thanks!

corob-msft:
Oops, I missed the part about 17.2. We'll fix it in post. Although I note the flag has been around a long time and already guarded the POSIX typedefs and defined macros.

Also, we recommend Microsoft people use the private repo, cpp-docs-pr, for their contributions, since it does additional build validations and also lets us embargo content for release.

docs/c-runtime-library/compatibility.md
```diff
@@ -20,7 +20,7 @@ The UCRT also implements a large subset of the POSIX.1 (ISO/IEC 9945-1:1996, the
...
+The C++ standard reserves names that begin with an underscore in the global namespace to the implementation. Both the POSIX functions and Microsoft-specific runtime library functions are in the global namespace, but aren't part of the standard C runtime library. That's why the preferred Microsoft implementations of these functions have a leading underscore. For portability, the UCRT also supports the default names, but the Microsoft C++ compiler issues a deprecation warning when code that uses them is compiled. Only the default names are deprecated, not the functions themselves. To suppress the warning, define `_CRT_NONSTDC_NO_WARNINGS` before including any headers in code that uses the original POSIX names. Because the C standard doesn't allow non-standard names in header files, by default [`/std:c11`](../build/reference/std-specify-language-standard-version.md) and [`/std:c17`](../build/reference/std-specify-language-standard-version.md) don't expose the default names for POSIX functions, types, and macros. If these names are necessary, define `_CRT_DECLARE_NONSTDC_NAMES` to expose them.
```
stwish-msft, line 23: > If these names are necessary, define `_CRT_DECLARE_NONSTDC_NAMES` to expose them.

This is actually an on/off switch. Defining that macro to 1 will enable the default names and defining that macro to 0 will disable the default names, regardless of the /std or /Za setting.

PR #3736, opened 2022-02-28
Typo

LittleDyingDuck:
Fixed some annoying typo that actually pissed me off severely. who would use an??

### DIFF
--- a/docs/mfc/relationship-between-a-cpp-window-object-and-an-hwnd.md
+++ b/docs/mfc/relationship-between-a-cpp-window-object-and-an-hwnd.md
@@ -1,12 +1,12 @@
 ---
-description: "Learn more about: Relationship Between a C++ Window Object and an HWND"
-title: "Relationship Between a C++ Window Object and an HWND"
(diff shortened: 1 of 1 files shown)


PRMerger5:
@LittleDyingDuck : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@LittleDyingDuck 
Thanks for proposing a change here, even though we're not going to adopt it. [The use of "a" or "an" depends on how you pronounce the word that follows it.](https://www.merriam-webster.com/words-at-play/is-it-a-or-an) If the pronunciation begins with a consonant sound, we use "a," and if it begins with a vowel sound, use "an." It's why we say "a one-hit wonder" an also "an hourly wage." Since the English pronunciation of "H" is "aitch," which starts with a vowel sound, and because that's the sound usually at the beginning of any pronunciation of `HWND` (at least among those of us who work at Microsoft), we use "an" as the indefinite article in front of it. When you pronounce `HWND`, you may use a different sound at the beginning, so it's possible that "a" would be correct for your usage. We'll stick with ours.

PR #3827, opened 2022-04-13
Update compiler-com-support.md

mrezam:
There is no need for extra notation of **microsoft specific** **end microsoft specific**. It seems to be a notation from an older documentation format which generates some unwanted bold texts in markdown document.


PRMerger-2:
@mrezam : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mrezam:
I have see the statements Microsoft specifc, END Microsoft specifc all over the place through the documentation. So if it the standard documentation format, the pull request is not valid anymore.

PR #3832, opened 2022-04-14
VMCS -> VMXON

krispybyte:
According to the intel IA32 development manual (Volume 3 C3 23.7):

"Before executing VMXON, software should allocate a naturally aligned 4-KByte region of memory that a logical
processor may use to support VMX operation. This region is called the **VMXON region**."

### DIFF
--- a/docs/intrinsics/vmx-on.md
+++ b/docs/intrinsics/vmx-on.md
@@ -1,7 +1,7 @@
 ---
 description: "Learn more about: __vmx_on"
 title: "__vmx_on"
-ms.date: "09/02/2019"
+ms.date: 04/14/2022"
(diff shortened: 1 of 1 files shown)


PRMerger7:
@krispybyte : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger9:

corob-msft (approved):
@krispybyte

Thanks for the update. While the original statement isn't wrong, your update is arguably more useful for finding out how to use the instruction. I've also updated the link to the relevant Intel docs, since the old search instructions don't return anything useful anymore. I'll touch the other articles that have this link, as well.

PR #3851, opened 2022-04-24
Fixes #3847 by using puts() and by NUL-terminating szMsg[]

### LINKED ISSUE #3847: Potential bugs/bad style in example for _getdiskfree
Hi, the example for _getdiskfree() bothered me in a couple of ways:

## Use of non-literal format string in printf()

There are several places where the example calls printf() with a single argument.  For example, "printf(g_szBorder);" or "printf(szMsg);".  This is generally bad style, because the argument could contain a "%", which would then cause printf() to try to consume arguments which aren't there.

Technically, this should be safe for cases such as "printf(g_szBorder);" since the string literals are defined above and do not contain a "%".  Even though it's okay, it's a "code smell" for me personally, since the literal is located in a different place than the call to printf(), so it requires a little bit of effort to verify.

The most troubling case is "printf(szMsg);" since in that case the format string comes from FormatMessage().  I don't know how likely it is that FormatMessage() would return a string containing a "%" character, but I'm not aware of any guarantee that it won't happen.

Generally, I would recommend doing either:

```c
printf("%s", str);
```

or

```c
fputs(str, stdout);
```

in order to make it obvious that there won't be any misbehavior if str contains "%".

## Removal of terminating NUL byte from result of FormatMessage()

In the error case, after calling FormatMessage(), the code does the following:

```c
            szMsg[uLen+6] = '  ';
            szMsg[uLen+7] = '  ';
            szMsg[uLen+8] = '  ';
```

My understanding is that 6 and 7 are to remove the "\r\n" at the end of the string, and 8 is to remove the terminating NUL from the end of the string.

This will work just fine, as long as the length of the string returned by FormatMessage() plus 8 fits within the length of g_szLine[], because then the NUL byte at the end of g_szLine[] will terminate the string.

However, the call to FormatMessage() specifies a maximum length of 4100 characters:

```c
            uLen = FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL,
                            uErr, 0, szMsg+8, 4100, NULL);
```

Therefore, if the message is long, FormatMessage() could easily overwrite the NUL byte at the end of the version of g_szLine[] which has been copied into szMsg[].  If that happens, there does not appear to be any guarantee that a NUL byte will be reached before reaching the end of szMsg[] when printing the message.  (Since szMsg[] is uninitialized, beyond the first sizeof(g_szLine) bytes.)

A simple fix would be to add:

```c
         memset(szMsg, 0, sizeof(szMsg));
```

before the existing

```c
         memcpy(szMsg, g_szLine, sizeof(g_szLine));
```

although there would be other ways to fix it, as well.

Thanks!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b3a07ec8-b631-20f0-c072-a03144c96487
* Version Independent ID: a49a780a-cb33-4ddd-8a7a-52bfea658e60
* Content: [_getdiskfree](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getdiskfree?view=msvc-170)
* Content Source: [docs/c-runtime-library/reference/getdiskfree.md](https://github.com/Microsoft/cpp-docs/blob/main/docs/c-runtime-library/reference/getdiskfree.md)
* Product: **visual-cpp**
* Technology: **cpp-ucrt**
* GitHub Login: @TylerMSFT
* Microsoft Alias: **twhitney**

ppelleti:
This is one potential way to solve the issues with the `_getdiskfree()` example I brought up in #3847.  There's more than one way to solve those problems, so if there's a different way you prefer better, that's fine.  I just wanted to suggest one potential way.

I changed `printf()` to `puts()`, since `printf()` was not being used to do any formatting.  Since `puts()` prints a newline after printing the string, I removed the newlines from the `g_szXXX[]` constants.

I added a check to see if the error message was longer than `g_szLine[]`, and if so, terminate the message with a NUL character, to avoid potentially reading past the end of the buffer.

**Note:** I don't currently have a Windows VM set up (it's on my to-do list) so I wasn't able to test my fixed version of the program.  So if you decide to go with my solution, please be sure to test it first, and make sure it compiles and runs.  Thanks!

### DIFF
--- a/docs/c-runtime-library/reference/getdiskfree.md
+++ b/docs/c-runtime-library/reference/getdiskfree.md
@@ -73,10 +73,10 @@ For more compatibility information, see [Compatibility](../../c-runtime-library/
 #include <stdio.h>
 #include <tchar.h>
 
-TCHAR   g_szBorder[] = _T("======================================================================\n");
-TCHAR   g_szTitle1[] = _T("|DRIVE|TOTAL CLUSTERS|AVAIL CLUSTERS|SECTORS / CLUSTER|BYTES / SECTOR|\n");
-TCHAR   g_szTitle2[] = _T("|=====|==============|==============|=================|==============|\n");
-TCHAR   g_szLine[]   = _T("|  A: |              |              |                 |              |\n");
+TCHAR   g_szBorder[] = _T("======================================================================");
+TCHAR   g_szTitle1[] = _T("|DRIVE|TOTAL CLUSTERS|AVAIL CLUSTERS|SECTORS / CLUSTER|BYTES / SECTOR|");
+TCHAR   g_szTitle2[] = _T("|=====|==============|==============|=================|==============|");
+TCHAR   g_szLine[]   = _T("|  A: |              |              |                 |              |");
 
 void utoiRightJustified(TCHAR* szLeft, TCHAR* szRight, unsigned uVal);
 
@@ -86,9 +86,9 @@ int main(int argc, char* argv[]) {
    ULONG uDriveMask = _getdrives();
    unsigned uErr, uLen, uDrive;
 
-   printf(g_szBorder);
-   printf(g_szTitle1);
-   printf(g_szTitle2);
+   puts(g_szBorder);
+   puts(g_szTitle1);
+   puts(g_szTitle2);
 
    for (uDrive=1; uDrive<=26; ++uDrive) {
       if (uDriveMask & 1) {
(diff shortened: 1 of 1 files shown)


PRMerger-2:
@ppelleti : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ShannonLeavitt:
@TylerMSFT – Can you review the proposed changes? IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

TylerMSFT (commented):
Hi @ppelleti,
Thank you for taking the time to update this example.
Something is wrong - I'm not getting the output I expect. Let me dig into that and I'll get back to you.

TylerMSFT:
I've fixed this by removing all the table formatting code, which just obscured what the topic was trying to demonstrate in the first place, which is _getdiskfree().

PR #3871, opened 2022-05-04
Update dynamicbase.md

laurenprinn:
arm32 and arm64 is not supported for this linker switch as ASLR cannot be disabled for these architectures.

### DIFF
--- a/docs/build/reference/dynamicbase.md
+++ b/docs/build/reference/dynamicbase.md
@@ -20,6 +20,8 @@ The **/DYNAMICBASE** option modifies the header of an *executable image*, a .dll
 
 By default, **/DYNAMICBASE** is enabled. To disable this option, use **/DYNAMICBASE:NO**. The **/DYNAMICBASE** option is required for the [/HIGHENTROPYVA](highentropyva-support-64-bit-aslr.md) option to have an effect.
 
+This option is not supported for Arm32 or Arm64 architectures.
+
 ## See also
 
 - [EDITBIN Options](editbin-options.md)
--- a/docs/build/reference/dynamicbase.md
+++ b/docs/build/reference/dynamicbase.md
@@ -1,26 +1,26 @@
 ---
 description: "Learn more about: /DYNAMICBASE"
 title: "/DYNAMICBASE"
-ms.date: "06/12/2018"
+ms.date: "05/05/2022"
 f1_keywords: ["/dynamicbase"]
 helpviewer_keywords: ["-DYNAMICBASE editbin option", "DYNAMICBASE editbin option", "/DYNAMICBASE editbin option"]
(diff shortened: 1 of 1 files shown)


PRMerger18:
@laurenprinn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@laurenprinn 
Thanks for the update. Should we make similar clarifications in the [`/HIGHENTROPYVA`](https://docs.microsoft.com/en-us/cpp/build/reference/highentropyva-support-64-bit-aslr?view=msvc-170) linker switch docs? When we mention ARM64, should we also include ARM64EC? And when you build for an ARM platform, do these flags get silently ignored, or do they cause command-line warning D9002 or similar?

corob-msft:
@laurenprinn 
A follow-up: Currently, you write "This option is not supported for Arm32 or Arm64 architectures." I'm thinking it might be more accurate to say "This option isn't available for ARM, ARM64, or ARM64EC targets because ASLR can't be disabled on these architectures." I'd substitute "is silently ignored" for "isn't available" if that's what actually happens. Is that a valid restatement of your intent?

Court72:
@corob-msft 

Thanks for reviewing the proposed changes. *IMPORTANT*: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

RussKeldorph:
> @laurenprinn A follow-up: Currently, you write "This option is not supported for Arm32 or Arm64 architectures." I'm thinking it might be more accurate to say "This option isn't available for ARM, ARM64, or ARM64EC targets because ASLR can't be disabled on these architectures." I'd substitute "is silently ignored" for "isn't available" if that's what actually happens. Is that a valid restatement of your intent?

@corob-msft You have the right idea. `/DYNAMICBASE` is actually supported just fine as it is the default for the ARM platforms.  It's actually `/DYNAMICBASE:NO` that is incompatible with them.  In some sense `/DYNAMICBASE` is silently ignored since it is the default and can't be overridden, but I don't know that we need to say that.  I would focus on `/DYNAMICBASE:NO`, and, yes, it's because the OS cannot disable ASLR on those platforms.

Regarding `/HIGHENTROPYVA`, the linker doesn't seem to enforce the same incompatibilities, so I'm not sure we need to change that.  Would need more investigation.

PRMerger7:

corob-msft (approved):

corob-msft:
Thanks, @laurenprinn  and @RussKeldorph. This looks good enough for posting. Let me know if you'd like to clarify anything else. I've put in a matching fix for the linker option (this one just touches the EDITBIN option docs) in [Followup cpp-docs 3871 in linker opts](https://github.com/MicrosoftDocs/cpp-docs-pr/pull/4262).

PR #3940, opened 2022-05-31
C++20 example was needed

rtischer8277:


ShannonLeavitt:
@TylerMSFT – Can you review the proposed changes? IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

docs/standard-library/allocators.md
```diff
@@ -18,6 +18,35 @@ class vector
 
 The C++ Standard Library provides a default implementation for an allocator. In C++11 and later, the default allocator is updated to expose a smaller interface; the new allocator is called a *minimal allocator*. In particular, the minimal allocator's `construct()` member supports move semantics, which can greatly improve performance. In most cases, this default allocator should be sufficient. In C++11 all the Standard Library types and functions that take an allocator type parameter support the minimal allocator interface, including `std::function`, `shared_ptr, allocate_shared()`, and `basic_string`.  For more information on the default allocator, see [`allocator` Class](allocator-class.md).
 
+## Writing Your Own Allocator (C++20)
+Several Allocator member functions have been removed: **`address`**, **`max_size`**, **`construct`** and **`destroy`**. **`allocate(n)`** creates array space of type T, but does not construct the objects. Possible exceptions are not handled in the minimal example below:
+
+```cpp
+include <iostream>
```
TylerMSFT, line 25: typo - missing #

docs/standard-library/allocators.md
```diff
@@ -18,6 +18,35 @@ class vector
...
+int main()
+{
+  // allocator for integer values
+  std::allocator<int> myAllocator;
+
+  // allocate space for five ints
+  int* arr = myAllocator.allocate( 5 );
+
+  // construct and assign ints to elements 0 and 3
+  arr[ 0 ] = 100;
```
TylerMSFT, line 37: our style is to not have spaces around array index, and to put the * with the variable, e.g. int *arr.  And we might as well name it what it is - myArray, or something.

docs/standard-library/allocators.md
```diff
@@ -18,6 +18,35 @@ class vector
...
+  arr[ 3 ] = 10;
+
+  std::cout << arr[ 0 ] << std::endl;
+  std::cout << arr[ 3 ] << std::endl;
+
+  // destruct elements 0 and 3 and deallocate space for the five ints
```
TylerMSFT, line 43: This comment is confusing to me (since ints aren't destructed). Maybe a more general statement that says // call destructors on the elements and deallocate the space

TylerMSFT (changes requested):
Thank you so much for adding to the docs! 
I left a couple comments.

docs/standard-library/allocators.md
```diff
@@ -18,6 +18,35 @@ class vector
 
 The C++ Standard Library provides a default implementation for an allocator. In C++11 and later, the default allocator is updated to expose a smaller interface; the new allocator is called a *minimal allocator*. In particular, the minimal allocator's `construct()` member supports move semantics, which can greatly improve performance. In most cases, this default allocator should be sufficient. In C++11 all the Standard Library types and functions that take an allocator type parameter support the minimal allocator interface, including `std::function`, `shared_ptr, allocate_shared()`, and `basic_string`.  For more information on the default allocator, see [`allocator` Class](allocator-class.md).
 
+## Writing Your Own Allocator (C++20)
```
StephanTLavavej, line 21: This added example is depicting usage of `std::allocator`; it is not depicting how a user would write their own allocator in C++20 (which is very similar to how it would be written in C++11, except for things like `operator!=` being unnecessary to provide).

docs/standard-library/allocators.md
```diff
@@ -18,6 +18,35 @@ class vector
...
+Several Allocator member functions have been removed: **`address`**, **`max_size`**, **`construct`** and **`destroy`**. **`allocate(n)`** creates array space of type T, but does not construct the objects. Possible exceptions are not handled in the minimal example below:
```
StephanTLavavej, line 22: > **`allocate(n)`** creates array space of type T, but does not construct the objects.

This has always been the case; it is not new to C++20.

docs/standard-library/allocators.md
StephanTLavavej: `deallocate()` doesn't call destructors at all (for any types); it expects the elements to have already been destroyed. In the case of `int` at runtime it doesn't really matter, which is why this example works.

docs/standard-library/allocators.md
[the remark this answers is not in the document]
StephanTLavavej: They were removed from `std::allocator`, but that is not especially relevant when talking about how to write a user-defined allocator.

StephanTLavavej (commented):
It is unclear to me whether this PR should be merged at all. Perhaps the "Writing Your Own Allocator (C++11)" section should be slightly amended to talk about C++20 changes.

rtischer8277:
I agree with the comments. My example code does not show how a user would write their own allocator. But the Mallocator example does and it doesn't use any of the C++20 deprecated methods **address**, **max_size**, **construct** or **destroy**. This means that my submitted code and text should be replaced with a simple note, as suggested, in the **Writing An Allocator (C++03)** section to the effect that, as of C++20 these 4 methods have been removed and that it is unnecessary to provide the **operator!=**.

PR #4027, opened 2022-07-06
Update walkthrough-creating-and-using-a-dynamic-link-library-cpp.md

garlfield:
Add description of precompiler steps in case the project name does not match.

### DIFF
--- a/docs/build/walkthrough-creating-and-using-a-dynamic-link-library-cpp.md
+++ b/docs/build/walkthrough-creating-and-using-a-dynamic-link-library-cpp.md
@@ -188,7 +188,7 @@ Right now, this DLL doesn't do very much. Next, you'll create a header file to d
 
 This header file declares some functions to produce a generalized Fibonacci sequence, given two initial values. A call to `fibonacci_init(1, 1)` generates the familiar Fibonacci number sequence.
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@garlfield : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
@corob-msft 

Can you review this PR?

IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

corob-msft:
@garlfield 
Thanks for being extremely patient. 

Looking at your addition, it's good to know how to set preprocessor definitions in your project properties. However, I'm having trouble with the idea that you'd ever need to change the `<PROJECTNAME>_EXPORTS` value in the middle of this walkthrough. If you create a new project with your preferred name, say, `MyNewLibrary`, that name will also be used in the preprocessor macros generated by the template, such as `MYNEWLIBRARY_EXPORTS`. You won't need to edit them after the fact. The only time you'd need the step you described is if you've decided to change the project name after creating the project, and you want everything to match - which is outside of the scope of the walkthrough.

It pains me to reject a submission, and the generous work it represents, but I'll need more convincing about its utility before I adopt this change. Feel free to make your case in comments below. For now, I'm closing the PR.

PR #4037, opened 2022-07-11
Update cpen-class.md

BigB9898:
0 pen width description is not clear.  Updating this document along with several others.

### DIFF
--- a/docs/mfc/reference/cpen-class.md
+++ b/docs/mfc/reference/cpen-class.md
@@ -177,7 +177,7 @@ Specifies the style for the pen. For a list of possible values, see the *`nPenSt
 *`nWidth`*\
 Specifies the width of the pen.
 
-- For the first version of `CreatePen`, a value of 0 will be treated similarly to a value of 1, except that the width will not be affected by scale-transform operations that are in effect for the Graphics object that the pen is used for, resulting in the width to be always 1 pixel..
+- For the first version of `CreatePen`, a value of 0 will be treated similarly to a value of 1, except that the width will not be affected by scale-transform operations that are in effect for the Graphics object that the pen is used for; the width will always be 1 pixel.
 
 - For the second version of `CreatePen`, if *`nPenStyle`* is `PS_GEOMETRIC`, the width is given in logical units. If *`nPenStyle`* is `PS_COSMETIC`, the width must be set to 1.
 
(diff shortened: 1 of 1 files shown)


PRMerger13:
@BigB9898 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TylerMSFT (changes requested):
Thank you, @BigB9898 . How you've rephrased this is better than the original, but I'm worried that we lost the specifics of the device units being 1 pixel. Could you strengthen that part? 
Super minor: the diff looks like there are now two periods at the end. That may be weirdness in how the diff is presented, but something to take a quick look at.

BigB9898:
> Thank you, @BigB9898 . How you've rephrased this is better than the original, but I'm worried that we lost the specifics of the device units being 1 pixel. Could you strengthen that part? Super minor: the diff looks like there are now two periods at the end. That may be weirdness in how the diff is presented, but something to take a quick look at.



> Thank you, @BigB9898 . How you've rephrased this is better than the original, but I'm worried that we lost the specifics of the device units being 1 pixel. Could you strengthen that part? Super minor: the diff looks like there are now two periods at the end. That may be weirdness in how the diff is presented, but something to take a quick look at.

Tyler, I updated per your feedback...Thank you.  Regarding the extra ".", when I previewed it was not there.

PRMerger12:
@BigB9898 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

BigB9898:
> Thank you, @BigB9898. This looks good. If you could fix the grammar, we'll be good to go. I'd put a period after "...that the pen is used for." and change "resulting in the width to be always 1 pixel." => "That is, the width will always be 1 pixel."

Thanks much!  Updated.

PRMerger7:
@BigB9898 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

BigB9898:
> We are almost there! Thank you very much for improving the docs. There is one instance of "resulting in the width to be always 1 pixel" If you make that match what you did on line 180, that would be much better and then we can merge this.

@TylerMSFT, thanks...Sorry I missed that update.

TylerMSFT (approved):
@BigB9898 , thank you for this update. It'll get merged and should hopefully be live by tomorrow.
Really appreciate you taking the time to do this. I know that you are doing it on your own time and just to file down a rough spot in the world to make it a little better place. Thank you.

TylerMSFT:
#sign-off

PR #4038, opened 2022-07-11
Code block changes

namanpatel6:
Submitted via Docs Editor.

### DIFF
--- a/docs/cpp/compound-statements-blocks.md
+++ b/docs/cpp/compound-statements-blocks.md
@@ -6,6 +6,7 @@ f1_keywords: ["}", "{"]
 helpviewer_keywords: ["blocks, about blocks", "compound statements"]
 ms.assetid: 23855939-7430-498e-8936-0c70055ea701
 ---
+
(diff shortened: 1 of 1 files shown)


PRMerger7:
@namanpatel6 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

corob-msft:
@namanpatel6 
We appreciate everyone who makes a contribution to improving the C++ docs, so, thank you. When you make a change, we like to see either a reference to an issue from our issues board, or a justification for the change in the comment for your PR. In this case, I'm having trouble understanding the reason for this change. Normally, you want to hoist common code paths outside of a control structure, so you don't repeat yourself (and so you don't make mistakes later when you make changes to one code path but not the other). If the `Alert` routine is meant to always be called, it should come after the block for the `else` clause. Since it only appears in the `if` clause, it's a clue that it only applies to the `if` (the error case) and not the `else` (the normal case). Is something unclear about this example?

Jak-MS:
#label:"aq-pr-triaged"

PR #4104, opened 2022-08-19
Fixed inconsistent spelling of "newline"

weijuwang:
"newline" was spelled as "new-line" two out of four times in the documentation.

### DIFF
--- a/docs/c-language/lexical-grammar.md
+++ b/docs/c-language/lexical-grammar.md
@@ -187,7 +187,7 @@ For a list of additional Microsoft-specific keywords, see [C keywords](./c-keywo
 &emsp;*`c-char-sequence`* *`c-char`*
 
 *`c-char`*:\
-&emsp;Any member of the source character set except the single quotation mark (**`'`**), backslash (**`\`**), or newline character\
+&emsp;Any member of the source character set except the single quotation mark (**`'`**), backslash (**`\`**), or new-line character\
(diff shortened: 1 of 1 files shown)


Jak-MS:
@corob-msft - Can you review this PR?

IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

corob-msft (changes requested):
@weijuwang 
Thanks for pointing out the inconsistency and proposing a fix. You have the right idea, but you've fixed the wrong targets. Ordinarily, I'd accept this PR, because the common term we use in the rest of the documentation is "newline". However, the [C lexical grammar](https://docs.microsoft.com/cpp/c-language/lexical-grammar?view=msvc-170) page quotes Annex A of the [C17 Standard](https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf), which uses "new-line" everywhere. To be consistent, we should change the two cases of "newline" to "new-line". I'll approve and merge the PR if you make those changes. Thank you again for taking the time to make the docs better.

weijuwang:
Should be fixed now. Thanks for the response!

corob-msft (approved):
Looks good to me. Thanks for the quick turnaround, @weijuwang

PR #4166, opened 2022-09-14
Fix markdown typo

apmorton:


corob-msft (approved):
@apmorton Looks good, thanks for the update.

PR #4167, opened 2022-09-15
Small correction in getch-getwch

NathanIsBored13:
'key code' implies virtual key code whereas the function actually returns a scan code on the second call, added link for further clarity


NathanIsBored13:
sorry for the corrections, this is my first time contributing and I messed up the link format

TylerMSFT:
@NathanIsBored13, this was a good addition. You were really close - just needed a leading '/'
I've made the change you intended to, and it should go live tomorrow.
Thank you for contributing to making the docs better!

PR #4176, opened 2022-09-20
Update clang-support-msbuild.md

walbourn:
Updated for better visibility of LLVMToolsVersion property.

Note the text was also misleading because it implies that LLVMToolsVersion supports side-by-side version of clang/LLVM in the same `LLVMInstallDir`, but that's not how clang/LLVM works. If you want to select a different version, you need to point it to a different path.

### DIFF
--- a/docs/build/clang-support-msbuild.md
+++ b/docs/build/clang-support-msbuild.md
@@ -1,7 +1,7 @@
 ---
 description: "Learn more about: Clang/LLVM support in Visual Studio projects"
 title: "Clang/LLVM support in Visual Studio projects"
-ms.date: 06/29/2022
+ms.date: 09/20/2022
 ms.description: "Configure a Visual Studio MSBuild project to use the Clang/LLVM toolchain."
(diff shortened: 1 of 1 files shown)


corob-msft:
@walbourn 
Do you think it makes sense to display different content for the VS2019 and VS2022 docs versions? I'm thinking specifically about the content on line 22. Someone might reasonably wonder "Well, should I use the v143 or the v142 build tools? You're not helping me choose."

jborsecnik:
@corob-msft, 

#label:"aq-pr-triaged"

walbourn:
> @walbourn Do you think it makes sense to display different content for the VS2019 and VS2022 docs versions? I'm thinking specifically about the content on line 22. Someone might reasonably wonder "Well, should I use the v143 or the v142 build tools? You're not helping me choose."

I was looking at that, but I don't know if the moniker tags can be nested, etc.

corob-msft (approved):
@walbourn 
The version question may not be significant. I patched up another area that seemed redundant after your changes.

PR #4202, opened 2022-10-04
Document ARM64 unwind handling for return address signing

mstorsjo:
This is deduced/guessed from studying the output of MSVC with the `/d2guardsignret` flag. Hopefully someone with insight into the unwinder internals can confirm whether this indeed is the case.

I made up an unwind opcode name for this opcode (0xfc, `sign_ra`) - other suggestions are welcome.

I'm not entirely sure about whether this is how it is handled for the packed functions, but this would intuitively be how it works.

Since there are two parallel instructions for signing return addresses, `paciasp` and `pacibsp`, the unwinder would need to know which one to use for signing - MSVC seems to generate `pacibsp` (togeter with `autibsp`) - is there any other similar opcode for `paciasp`, or is that one unsupported for Windows unwinding purposes?

### DIFF
--- a/docs/build/arm64-exception-handling.md
+++ b/docs/build/arm64-exception-handling.md
@@ -1,7 +1,7 @@
 ---
 title: "ARM64 exception handling"
 description: Describes the exception handling conventions and data used by windows on ARM64.
-ms.date: 04/07/2022
+ms.date: 10/10/2022
 ---
 # ARM64 exception handling
 
@@ -344,7 +344,7 @@ The fields are as follows:
 - **Function Length** is an 11-bit field providing the length of the entire function in bytes, divided by 4. If the function is larger than 8k, a full `.xdata` record must be used instead.
 - **Frame Size** is a 9-bit field indicating the number of bytes of stack that is allocated for this function, divided by 16. Functions that allocate greater than (8k-16) bytes of stack must use a full `.xdata` record. It includes the local variable area, outgoing parameter area, callee-saved Int and FP area, and home parameter area. It excludes the dynamic allocation area.
 - **CR** is a 2-bit flag indicating whether the function includes extra instructions to set up a frame chain and return link:
-  - 00 = unchained function, \<x29,lr> pair isn't saved in stack.
+  - 00 = unchained function, \<x29,lr> pair isn't saved in stack
   - 01 = unchained function, \<lr> is saved in stack
   - 10 = chained function with signed return address
   - 11 = chained function, a store/load pair instruction is used in prolog/epilog \<x29,lr>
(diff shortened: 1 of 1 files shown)


corob-msft:
@mstorsjo 

Thanks for the proposal. As near as I can tell, the actual opcode is known as `pac_sign_return_address`, and the logic is largely as you've divined. There appears to be a flag that lets the signing/unsigning opcodes be replaced with nops, so there's an additional bit of complexity involved somewhere. At least, I think so, if I'm reading things correctly, which is always a question. The signing step is "Step 1" in the internal comments, and saving registers is "Step 2", so we should probably renumber those and the following steps accordingly.
 
To get the inside story, I'm invoking @laurenprinn and @sigatrev for their takes. (I'd invoke James McNellis, but he's moved on to smaller and more playful things at Roblox.) I don't feel qualified to sign off on my own.

mstorsjo:
> Thanks for the proposal. As near as I can tell, the actual opcode is known as `pac_sign_return_address`, and the logic is largely as you've divined. There appears to be a flag that lets the signing/verifying opcodes be replaced with nops, so there's an additional bit of complexity involved somewhere. At least, I think so, if I'm reading things correctly, which is always a question. The signing step is "Step 1" in the internal comments, and saving registers is "Step 2", so we should probably renumber those and the following steps accordingly.

Thanks! I've updated the PR to use the correct name for the opcode, and renumbered the steps for the packed form.

docs/build/arm64-exception-handling.md
```diff
@@ -313,6 +313,7 @@ The unwind codes are encoded according to the table below. All unwind codes are
 |  | 11101010: Custom stack for MSFT_OP_CONTEXT |
 |  | 11101100: Custom stack for MSFT_OP_CLEAR_UNWOUND_TO_CALL |
 |  | 1111xxxx: reserved |
+| `pac_sign_return_address` | 11111100: sign the return address in `x30` with `pacibsp` |
```
sigatrev, line 316: `x30` is referred to as `lr` everywhere else in this document. We should use `lr` here was well for consistency.

mstorsjo: Thanks - I updated the branch, fixing this issue now, so that it references `lr` instead.

sigatrev (approved):

mstorsjo:
@sigatrev BTW, are you able to comment on what the Windows unwinder implementation does, if the actual code does use PAC (and running on HW that exposes it, with a Windows configuration that has taken it into use), but the unwind info is lacking these opcodes? (That's the current status quo if building code with Clang today, with `-mbranch-protection=standard`.)

pmsjt:
> @sigatrev BTW, are you able to comment on what the Windows unwinder implementation does, if the actual code does use PAC (and running on HW that exposes it, with a Windows configuration that has taken it into use), but the unwind info is lacking these opcodes? (That's the current status quo if building code with Clang today, with `-mbranch-protection=standard`.) I.e., does the unwinder implicitly run `xpaci` on the return address after finishing `RtlVirtualUnwind`, or does it only do that (or similar) if explicitly told so by an unwinding opcode?

It must be explicitly present.

mstorsjo:
> > @sigatrev BTW, are you able to comment on what the Windows unwinder implementation does, if the actual code does use PAC (and running on HW that exposes it, with a Windows configuration that has taken it into use), but the unwind info is lacking these opcodes? (That's the current status quo if building code with Clang today, with `-mbranch-protection=standard`.) I.e., does the unwinder implicitly run `xpaci` on the return address after finishing `RtlVirtualUnwind`, or does it only do that (or similar) if explicitly told so by an unwinding opcode?
> 
> It must be explicitly present.

Ok, thanks for the clarification!

corob-msft (approved):
@mstorsjo 
Thanks for the useful update.

corob-msft:
#sign-off

PR #4209, opened 2022-10-11
Rename pac_sign_lr and update reserved ranges.

pmsjt:


Court72:
@corob-msft 

#label:"aq-pr-triaged"

corob-msft:
@pmsjt 
Thanks for the expansion. While the compact opcode name appears to be arbitrary, and isn't documented anywhere else, is there a reason other than terseness for choosing `pac_sign_lr`? At least one internal tool appears to dump the compact opcode as "PAC_SIGN_RETURN_ADDRESS". The linker dumps the compact opcodes as their regular opcode equivalents, so it doesn't get exposed that way.

pmsjt:
> @pmsjt Thanks for the expansion. While the compact opcode name appears to be arbitrary, and isn't documented anywhere else, is there a reason other than terseness for choosing `pac_sign_lr`? At least one internal tool appears to dump the compact opcode as "PAC_SIGN_RETURN_ADDRESS". The linker dumps the compact opcodes as their regular opcode equivalents, so it doesn't get exposed that way.

Unwind opcode names are usually short. The previous name stood out like a sore thumb.

corob-msft (approved):

PR #4210, opened 2022-10-11
example needed to mitigate dev floundering

rtischer8277:
It is not easy to match up variant vt types with POD types. Especially since there is a gradual transition to modern C++. I used a number of dev hours on attempts to match the variant type, vt, to make the true/false work. The m_bValue type for the simple property was a BOOL  type, but trying to create a variant of type BOOL directly didn't work. Only by looking at VC Sample code for VisualStudioDemo did I see the (_variant_t)true construct, which becomes the standard_cast contruct in modern C++. I figured my Example code could make other programmer's efforts easier.

### DIFF
--- a/docs/mfc/reference/cmfcpropertygridproperty-class.md
+++ b/docs/mfc/reference/cmfcpropertygridproperty-class.md
@@ -1513,7 +1513,17 @@ Sets the value of a property grid property.
 ```
 virtual void SetValue(const _variant_t& varValue);
 ```
-
+Example:
+```
+void SetPropBarValue( UINT propId, const DWORD& barPropDwordValue )
+{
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT (approved):
@rtischer8277 , thank you very much for adding to the docs! We really appreciate that you'd take the time to add your learnings to the doc. There are some minor stylistic things in terms of spacing, and we prefer to say if (property != null) instead of if (property). But these are small things and I'll update them later. The important thing is the example you contributed and I'm approving your changes so we can get them in and published tomorrow. Thank you!

TylerMSFT:
#sign-off

PR #4214, opened 2022-10-12
Use the correct canonical opcode name in the packed unwind table

mstorsjo:
This was a leftover from the original tentative name before the real one was known.

Sorry for missing this...


corob-msft (changes requested):
@mstorsjo 
Since the name is arbitrary, @pmsjt (whose fingerprints are on the Windows ARM64 unwind code) convinced me to shorten it to `pac_sign_lr` instead. I couldn't find any evidence the longer name is exposed anywhere except in a tool internal to Microsoft.

mstorsjo:
> Since the name is arbitrary, @pmsjt (whose fingerprints are on the Windows ARM64 unwind code) convinced me to shorten it to `pac_sign_lr` instead. I couldn't find any evidence the longer name is exposed anywhere except in a tool internal to Microsoft.

Ok, great, thanks! Yeah a shorter name would be convenient.

Although, I had just landed a change in LLVM making this name used in an external assembly directive name, in https://github.com/llvm/llvm-project/commit/c43bff64e903f726d456ca27fd167b91198ae169. But since it literally landed a couple hours ago (and we're months away from a release), it's shouldn't be a problem to switch it over there to a less unwieldy name (and that's certainly appreciated).

pmsjt:
Looks good to me. Thanks.

corob-msft (approved):

PR #4224, opened 2022-10-14
Update profile-guided-optimizations.md

ArminG-MSFT:
ARM64 is also supported.


corob-msft:
@ArminG-MSFT
Thanks for expanding on that. Do you know when ARM64 support was added? As in, which version of Visual Studio first supported ARM64 POGO? If this is a more recent addition than ARM64 support in general, it's useful to let readers know they need to update their VS to get the feature.

corob-msft (approved):
@ArminG-MSFT Sounds like versions aren't a worry, so LGTM, thanks!

corob-msft:
Re-opening to trigger the CLA check, which should pass for all members of Microsoft.

PR #4234, opened 2022-10-19
fopen-wfopen.md: initial comma for ccs= option

skissane:
Unlike the other mode options, `ccs=` must always be prefixed with a comma. The examples made this clear, but the table and text didn't. I think that point is made more obvious if one always writes `,ccs=`

### DIFF
--- a/docs/c-runtime-library/reference/fopen-wfopen.md
+++ b/docs/c-runtime-library/reference/fopen-wfopen.md
@@ -53,7 +53,7 @@ By default, this function's global state is scoped to the application. To change
 
 ## Unicode support
 
-**`fopen`** supports Unicode file streams. To open a Unicode file, pass a `ccs=encoding` flag that specifies the desired encoding to **`fopen`**, as follows.
+**`fopen`** supports Unicode file streams. To open a Unicode file, pass a `,ccs=encoding` flag that specifies the desired encoding to **`fopen`**, as follows.
 
 `FILE *fp = fopen("newfile.txt", "rt+, ccs=UTF-8");`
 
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT:
Hi @skissane, 
Thank you for taking an interest in the docs.
The comma isn't part of the flag. It's there because the ccs flag is following another argument. Although it needs to follow the mode argument (meaning it'll usually be separated from the previous mode argument by a comma), that doesn't mean the comma is part of the flag. It's just part of following another parameter. It would be confusing if we suggested in the docs that argument delimiters were part of the definition of the flag when that flag has some sort of positional requirement because it's not part of the flag--it's just normal parameter passing.
It's awesome of you to notice something that strikes you funny and take the time to make a contribution. In this case, I don't think we will include it in the official docs, though.

skissane:
@TylerMSFT I take your point point about the comma not actually being part of the flag.

I do still think there is an issue here though. `fopen()` mode flags fall into two groups:
(1) firstly, single character flags
(2) subsequently, comma-separated keyword flags, such as `ccs`

Only (1) is standardised as part of the ISO C standard, although the [C99 rationale document](https://www.open-std.org/jtc1/sc22/wg14/www/C99RationaleV5.10.pdf) (see section 7.19.5.3, page numbered 150, PDF page 157) gives an example of (2) as a possible extension. For Windows UCRT, `ccs` is the only supported comma-separated keyword flag; likewise, for GNU C Library, it is the only supported comma-separated flag. However, on some other platforms (for example, [IBM z/OS](https://www.ibm.com/docs/en/zos/2.1.0?topic=functions-fopen-open-file)), there are many more comma-separated keyword flags.

Anyway, my point is, the documentation text never clearly explains that `ccs` has a rather different syntax from all the other flags. On the contrary, it puts it in the same table with all the others. The examples in the docs do show this, but I think it would be easier for people to understand if it was explicitly stated in the text.

Maybe my proposed change was not the best way of communicating this, but I do think the documentation would be more understandable if this fact was made explicit, *somehow*.

TylerMSFT:
@skissane , I addressed your point by adding the following to the css entry in the table: This flag is separated from flags that precede it by a comma (`,`). For example: `FILE *f = fopen("newfile.txt", "rt+, ccs=UTF-8");`  It should be live by end of day Thursday (12/22) Happy holidays.

PR #4260, opened 2022-11-04
Update cpp-type-system-modern-cpp.md

drvoss:
Add semicolon


corob-msft (approved):
@drvoss Looks like a worthy addition to the example. Thanks!

PR #4303, opened 2022-11-25
system-function.md: remove outdated language

skissane:
This text talks about "Windows" and "Windows NT" as if they were two different things. It looks like the text dates back to the Windows 9x/Me era, when that was true. But, that hasn't been true for over 20 years now. Trying to update it.


jborsecnik:
@corob-msft, 

#label:"aq-pr-triaged"

corob-msft (approved):
@skissane Thanks for the update. Much of the documentation that refers to ANSI C is in fact nearly thirty years old, and sometimes older, so I'd guess in this case that the distinction made between Windows and Windows NT dates back to the Windows 3.1x era when Windows NT was new.

PR #4313, opened 2022-11-30
Remove definite article from "The AddressSanitizer"

Porges:
This is a proper noun and it reads as if we wrote "the Visual Studio" everywhere. There are probably other documents to update as well.

### DIFF
--- a/docs/sanitizers/asan.md
+++ b/docs/sanitizers/asan.md
@@ -10,7 +10,7 @@ helpviewer_keywords: ["ASan", "AddressSanitizer", "Address Sanitizer", "compilin
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


corob-msft (approved):
@Porges Thanks for the updates. I fiddled with them slightly to make our grammar checker happy.

docs/sanitizers/asan.md
```diff
@@ -34,17 +34,17 @@ Use AddressSanitizer to reduce your time spent on:
...
 
 - Local machine testing,
 - On-premise distributed testing, and
 - Cloud-based workflows for testing.
 
-### Install the AddressSanitizer
+### Install AddressSanitizer
 
-The AddressSanitizer IDE integration and libraries get installed by default with C++ workloads in the Visual Studio Installer. However, if you're upgrading from an older version of Visual Studio 2019, use the Installer to enable ASan support after the upgrade:
+C++ workloads in the Visual Studio Installer install AddressSanitizer IDE integration and libraries by default. However, if you're upgrading from an older version of Visual Studio 2019, use the Installer to enable ASan support after the upgrade:
```
Porges, line 47: @corob-msft I think the `the` here should have been retained

corob-msft: Yes, I expect so too.

corob-msft: A fix will appear Monday.

PR #4338, opened 2022-12-13
update clarity

molecula451:
we can make this one clearer, usually we seen <code> extern 'C' ...</code> in header files, & avoid paraphrasing redundancy in "source code file"

### DIFF
--- a/docs/cpp/const-cpp.md
+++ b/docs/cpp/const-cpp.md
@@ -165,7 +165,7 @@ But to get the same behavior in C++, you must declare your **`const`** variable 
 extern const int i = 2;
 ```
 
-If you wish to declare an **`extern`** variable in a C++ source code file for use in a C source code file, use:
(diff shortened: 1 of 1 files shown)


Court72:
@corob-msft 

#label:"aq-pr-triaged"

corob-msft:
@molecula451 

I appreciate the time and effort you've put into improving the documentation. Unfortunately, this change isn't appropriate here. It's true that you can declare an `extern "C"` variable in a header file, but you can't include that header in a C source file, since C doesn't know what `extern "C"` means. Similarly, you don't want to define an `extern "C"` variable in a header file (that is, initialize it), because that can lead to multiple definitions that will make the linker throw an error. The example in this article really is meant to be used in a C++ source file (the sort with a *`.cpp`* extension). So, I'm sorry your effort isn't being rewarded, but this PR shouldn't be merged.

molecula451:
my bad, re-read the stuff, i tought i was read 2 times, cpp, thanks all good

PR #4354, opened 2023-01-04
Update integritycheck-require-signature-check.md

jsuther1974:


corob-msft (approved):
@jsuther1974 Thanks for the update.

PR #4377, opened 2023-01-20
fix typo - change "map" to "set" on <set> functions page

donkervoet6:
The <set> functions page lists "swap (map)", which should actually be "swap (set)"


TylerMSFT (approved):
Thank you for taking the time to fix this! Much appreciated.

PR #4378, opened 2023-01-20
BitScanForward clarification to avoid ambiguous "return" term.

mstrgram:
The remarks section has been reworded to avoid confusing use of the word "return" for both actual return value and the bit index written to the address given in the first parameter.


corob-msft (approved):
@mstrgram Thanks for the clarification.

PR #4408, opened 2023-02-08
remove `\r` from printf statements

strega-nil-ms:
`printf` already adds a `\r` when printing to the console, or writing text files.


corob-msft (approved):
@strega-nil-ms Thanks for the update!

PR #4411, opened 2023-02-10
Update SECURITY.md

ErikEast94:


corob-msft:
@ErikEast94 

I'm sorry, this file and others at the top level of the public repo can't be edited in cpp-docs. Only members of the appropriate Microsoft team may change these files. Thanks for your understanding.

PR #4415, opened 2023-02-14
Format /Fopathname so it doesn't get separated

snakecase:
Browsers render **`/Fo`***`pathname`* as two separate strings which confuses those who want to quickly glance at the syntax and move on. "Remarks" section clearly says "Don't put a space between the /Fo option and the pathname argument".

### DIFF
--- a/docs/build/reference/fo-object-file-name.md
+++ b/docs/build/reference/fo-object-file-name.md
@@ -6,13 +6,13 @@ f1_keywords: ["/Fo", "VC.Project.VCCLCompilerTool.ObjectFile", "VC.Project.VCCLW
 helpviewer_keywords: ["Fo compiler option [C++]", "object files, naming", "/Fo compiler option [C++]", "-Fo compiler option [C++]"]
 ms.assetid: 0e6d593e-4e7f-4990-9e6e-92e1dcbcf6e6
 ---
-# `/Fo` (Object File Name)
+# `/Fo`(Object File Name)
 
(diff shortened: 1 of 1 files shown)


Court72:
@corob-msft

#label:"aq-pr-triaged"

corob-msft (commented):
@snakecase 
I appreciate your initiative and commend you for looking out for your fellow docs users. Your intentions are good, but they've run into an issue we've argued about with the team in charge of our CSS styling for years. The title doesn't need to be changed, since the part in parentheses is the description of the option, not the description of the argument to the option. The syntax block, on the other hand, isn't supposed to look like it has a space between the option and the pathname. The only reason it does is because the CSS for code styling has some unnecessary horizontal padding, and we have to apply the bold code styling and the italicized code styling in separate sections. We've opted to stick with the standardized styling and live with the apparent space, making sure to comment that it isn't supposed to be there. It's an imperfect solution, but for consistency across the docs it's what we're sticking with for now. My apologies for not accepting your PR.

corob-msft:
Closing as not consistent with our style.

snakecase:
Sad that consistency (especially where it clearly works against the user) takes precedence over common sense. This PR was made only because a few people on my team have already lost some time trying to figure out why the command doesn't work and were really confused about documentation confirming their expectations of `/command argument` format.

I propose to put the `Don't put a space between the /Fo option and the pathname argument` as the first sentence at least.

PR #4456, opened 2023-03-07
masm/bnf-grammar: Format some literal characters correctly

lkundrak:
A couple of character literal terminals don't follow consistent formatting. Fix that.


TylerMSFT (approved):
I'm on the fence about the value of these changes. One of the reasons we do code escape is to prevent machine translation of identifiers. But that's not an issue with parenthesis, etc.
I'll take the change since the work is done, but I wouldn't necessarily encourage further work like this unless it is to prevent machine translation from incorrectly modifying a code statement.

TylerMSFT:
#sign-off

TylerMSFT (approved):
Hi, thanks for the changes.

PR #4464, opened 2023-03-15
Update 'Connect to your target Linux system in Visual Studio'.

fkelava:
Clearly state that Windows' default `ssh-keygen` does not produce valid private keys for use with ConnectionManager/VS, including steps to adjust private keys to conform with it.

Specifically, the reasoning behind this change _and_ the underlying issue are explained clearly in the following Developer Community tickets and SO threads:
https://developercommunity.visualstudio.com/t/Connect-to-Remote-System-fails-to-esta/10311053
https://developercommunity.visualstudio.com/t/cannot-connect-to-remote-using-private-key/1555749
https://stackoverflow.com/questions/53134212/invalid-privatekey-when-using-jsch

Until such a time as ConnectionManager is fixed to be compliant with default `ssh-keygen` options in modern Windows editions, it should be stated clearly that this is not the case and how this can be resolved.

### DIFF
--- a/docs/linux/connect-to-your-remote-linux-computer.md
+++ b/docs/linux/connect-to-your-remote-linux-computer.md
@@ -83,6 +83,9 @@ If `ssh` isn't already set up and running on your Linux system, follow these ste
    | **Passphrase** | Passphrase used with private key selected above |
 
    You can use either a password or a key file and passphrase for authentication. For many development scenarios, password authentication is sufficient, but key files are more secure. If you already have a key pair, it's possible to reuse it. Currently Visual Studio only supports RSA and DSA keys for remote connections.
+   
+   > [!NOTE]
+   > If using `ssh-keygen` to create the private key, you must specify the switch `-m pem`, or the key will not be accepted by Visual Studio. If your private key begins with `-----BEGIN OPENSSH PRIVATE KEY-----`, you must convert it with `ssh-keygen -p -f <FILE> -m pem`. 
 
 1. Choose the **Connect** button to attempt a connection to the remote computer.
 


Court72:
@tylermsft

#label:"aq-pr-triaged"

fkelava:
To summarize the issue more neatly:

Visual Studio's `ConnectionManager.exe` component that handles, among others, SSH connections still only accepts private keys in "old" OpenSSH format (today available from `ssh-keygen` with the option `-m pem`). The old format has been superseded in OpenSSH 7.8, released in [August of 2018](https://www.openssh.com/txt/release-7.8). See the first entry under 'Potentially-incompatible changes'.

The problem is not so much the fact that `ConnectionManager` rejects these keys as the fact Visual Studio does not make this clear in any way. In Visual Studio, if you are trying to add a SSH connection through the 'Connect to Remote System' window, you will instead see a message of `'Failed to negotiate host key algorithm. Only RSA and DSA host keys are supported.'`, which is in this case entirely incorrect. The article I seek to edit talks at length about VS' supported ciphers and key/kex/MAC algorithms starting from v16.9, but does _not_ mention this private key requirement.

It _is_ mentioned in [Set up FIPS-compliant secure remote Linux development](https://learn.microsoft.com/en-us/cpp/linux/set-up-fips-compliant-secure-remote-linux-development?view=msvc-170), but I find it unintuitive that a general requirement for _all_ private keys should be found in an article concerning FIPS compliance.

Making this clear in the documentation is merely a stopgap, but the issue should be properly resolved by finally updating `ConnectionManager` to be capable of handling keys OpenSSH produces _by default_ for `-t rsa` for a solid four and a half years now.

TylerMSFT (approved):
@fkelava , thank you for adding this to the docs! It should be live by end of day Monday.

TylerMSFT:
#sign-off

xPaw:
Well I am currently ratholing into this issue and using `-m pem` is not actually converting into a PEM key (nor does it generate a new one in this format).

fkelava:
If I haven't misread the man page, it seems the equivalent command is now ``ssh-keygen -e -f <FILE> -m pem``. Does that work?

If so, I can amend the docs again.

xPaw:
`do_convert_to_pem: unsupported key type ED25519` seems not.

I did get it to eventually work by making a new key, but it was pointless because debugging a self contained .NET binary remotely doesn't work anyway.

PR #4527, opened 2023-04-25
Correct documentation for HFA return values

sigatrev:
The listed rules for HVA return values was correct, but for HFAs it matters whether or not the type is an aggregate.

### DIFF
--- a/docs/build/arm64-windows-abi-conventions.md
+++ b/docs/build/arm64-windows-abi-conventions.md
@@ -204,7 +204,7 @@ A type is considered to be an HFA or HVA if all of the following hold:
 - It doesn't have any non-trivial default or copy constructors, destructors, or assignment operators,
 - All of its members have the same HFA or HVA type, or are float, double, or neon types that match the other members' HFA or HVA types.
 
-HFA and HVA values with four or fewer elements are returned in s0-s3, d0-d3, or v0-v3, as appropriate.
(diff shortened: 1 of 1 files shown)


TylerMSFT (approved):
Thanks, Matt.

TylerMSFT:
#sign-off

docs/build/arm64-windows-abi-conventions.md
```diff
@@ -204,7 +204,7 @@ A type is considered to be an HFA or HVA if all of the following hold:
 - It doesn't have any non-trivial default or copy constructors, destructors, or assignment operators,
 - All of its members have the same HFA or HVA type, or are float, double, or neon types that match the other members' HFA or HVA types.
 
-HFA and HVA values with four or fewer elements are returned in s0-s3, d0-d3, or v0-v3, as appropriate.
+HVA values with four or fewer elements are returned in s0-s3, d0-d3, or v0-v3, as appropriate.
 
 Types returned by value are handled differently depending on whether they have certain properties, and whether the function is a non-static member function. Types which have all of these properties,
```
efriedma-quic, line 209: Maybe clearer to explicitly state "non-HVA types" here?

efriedma-quic: Actually, nevermind; there are rules for both HVAs and HFAs that aren't stated here, so that fix doesn't really help.  I'm trying to work out the exact rules...

PR #4547, opened 2023-05-14
Clarify terminology around register volatility.

pmsjt:
Volatile vs Non-volatile is a very Microsoft specific terminology. Out in the industry, register calling convention assignment are more frequently referred to as Caller vs Callee saved.

Also, x18 is a reserved register, so it is neither Caller nor Callee Saved.

### DIFF
--- a/docs/build/arm64-windows-abi-conventions.md
+++ b/docs/build/arm64-windows-abi-conventions.md
@@ -66,16 +66,16 @@ Default layout alignment for globals and statics:
 
 The AArch64 architecture supports 32 integer registers:
 
-| Register | Save | Role |
+| Register | Volatility | Role |
 | - | - | - |
-| x0 | Caller | Parameter/scratch register 1, result register |
-| x1-x7 | Caller | Parameter/scratch register 2-8 |
-| x8-x15 | Caller | Scratch registers |
-| x16-x17 | Caller | Intra-procedure-call scratch registers |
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (changes requested):
Hi @pmsjt, it was cool of you to update this page. I'm unfortunately not going to take this PR because although I think you are right that caller-saved and callee-saved are more widely used outside of MS, these changes then make this one doc inconsistent with our other docs. For better or for worse, MS went down the road of this terminology and I'd want to make a MS-wide change to our documentation rather than become inconsistent with ourselves. It's not a wonderful situation but I unfortunately don't have the time to undertake such an effort.

pmsjt:
@TylerMSFT please reopen the PR so I can address the feedback. I don't have the permissions to do so. If you Close the PR I can't address the comments.

pmsjt:
Thank you for the feedback @TylerMSFT. I agree it is best to keep terminology consistent. I revised the change to touch only the cases where taxonomy was objectively incorrect and aimed to clarify the usage without introducing new terminology.

Please let me know if this is more palatable.

TylerMSFT (approved):
Thank you very much!

Court72:
@TylerMSFT

Thanks for reviewing the proposed changes.

TylerMSFT:
#sign-off

PR #4616, opened 2023-07-04
Clarified which New Project should be selected (for VS 2022)

enricosebastian:
There are two options; one does not work (.NET) due to the prerequisites set in the guide.

### DIFF
--- a/docs/dotnet/walkthrough-compiling-a-cpp-program-that-targets-the-clr-in-visual-studio.md
+++ b/docs/dotnet/walkthrough-compiling-a-cpp-program-that-targets-the-clr-in-visual-studio.md
@@ -26,7 +26,7 @@ The following steps vary depending on which version of Visual Studio you are usi
 
 1. In **Solution Explorer**, right-click on the top  to open the **Create a New Project** dialog box.
 
-1. At the top of the dialog, type **CLR** in the search box and then choose **CLR Empty Project** from the results list.
+1. At the top of the dialog, type **CLR** in the search box and then choose **CLR Empty Project (.NET Framework)** from the results list.
 
 1. Choose the **Create** button to create the project.
 


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
Thanks, @enricosebastian.  I think the problem is deeper in this topic because it doesn't discuss at all the difference between creating a project for .NET Framework (the original Windows-only version of .NET) or what was once known as .NET Core (the .NET option). You mentioned the pre-requisites pushing you to needing the windows version. I don't see that in the prereqs, though if you do try the suggested example then it needs to be the windows version (.NET Framework).  Can you tell me a little more about what in the prereqs suggested to you that it needed to be the Windows version of the .NET framework?  I agree a change needs to be made here, I think it may be more than just picking a different template.

enricosebastian:
> Thanks, @enricosebastian. I think the problem is deeper in this topic because it doesn't discuss at all the difference between creating a project for .NET Framework (the original Windows-only version of .NET) or what was once known as .NET Core (the .NET option). You mentioned the pre-requisites pushing you to needing the windows version. I don't see that in the prereqs, though if you do try the suggested example then it needs to be the windows version (.NET Framework). Can you tell me a little more about what in the prereqs suggested to you that it needed to be the Windows version of the .NET framework? I agree a change needs to be made here, I think it may be more than just picking a different template.

Hi @TylerMSFT! Been following the C++/CLI documentation for a few days now because I will be needing it for a project at work. Followed the guide, but found out that there would be two CLR Empty Projects options; one that's .NET and another that's .NET Framework. It didn't specify nor differentiate between the two in the walkthrough (or perhaps I missed that par. Sorry, rushing my project hahaha).

Decided to try both .NET and .NET Framework; the former would result in `...\...\...\x64\Debug\[name].dll is not a valid Win32 application`, but the .NET Framework would work with no hiccups (and is the method I am currently using for work). Did not decide to look further into the issue yet, but I did follow the setup prereqs via VS Installer.

I do agree with you that a total rework might be needed (or at least the section is split to differentiate between the two C++/CLI projects), but right now this edit will be a big help for those like me just starting to read through the documentation, and just need to see the output

TylerMSFT (approved):
Thanks, @enricosebastian!

TylerMSFT:
#sign-off

PR #4641, opened 2023-07-21
Suggestion for a better example program

MAP233224:
The format string in the printf_s explains what __umulh does better, visually speaking, by zero-padding all 64-bit values and separating the high and low parts of the product.


TylerMSFT (approved):
Thank you @MAP233224. This will be live by Monday.

TylerMSFT:
#sign-off

PR #4658, opened 2023-07-27
largeaddressaware:no is not recommended for 64-bit apps

mspll:


Jak-MS:
@TylerMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

TylerMSFT (changes requested):
Hi @mspll,
Thank you for taking the time to contribute to the docs!
I'd like to take this contribution but would like to tone down the 'result in runtime failures' part. The way it is written implies that using the switch can generate runtime failures. As you know, the problem is that memory usage is limited to 2gb, and if your program fills it up (which most don't come close to) then you could potentially hit a runtime error. But I don't want to imply that the switch results in runtime failures. Can you update?

mspll:
@TylerMSFT One case where this causes runtime failures is trying to run a non-large-address-aware x64 EXE on an ARM64 system. The emulation runtime tries to reserve a 4 GB region of VA space and fails, so the process doesn't even start.

GitTyler:
That's an interesting scenario. What do you think about something like this, then:

Linking 64-bit applications with **`/LARGEADDRESSAWARE:NO`** is not recommended. It restricts the amount of available address space available to your app which can result in runtime failures if the app exhausts memory. It may also prevent x64 apps from running on ARM64 systems because the emulation runtime will try to reserve 4GB of virtual address space. If the app was linked with `/LARGEADRESSAWARE:NO`, the app won't launch because allocating that much address space will fail.

Awesome of you to update this page as a result of your experience!

TylerMSFT (approved):

TylerMSFT:
#sign-off

PR #4676, opened 2023-08-09
[link.exe /INTEGRITYCHECK] fix -- ":NO" suffix doesn't work, so update the docs

glenn-slayden:
The doc mentions `/INTEGRITYCHECK:NO`, but this does not appear to be a valid command line option with the current `link.exe` releases. My version of `link` rejects that syntax entirely.

I don't know where the official `link.exe` source code is kept, but it likely has the exact same bug in the code that I found in the link below. Due to this obvious bug that I point out in the source code, there seems to be no way to explicilty assert `/INTEGRITYCHECK` in the negative, except by simply removing the option. 

https://github.com/microsoft/BuildXL/blob/main/Public/Sdk/Experimental/Msvc/Native/Tools/Link/Link.dsc#L303-L307

I would welcome a `link` developer to fix the program itself, so it works as described, but in the meantime, at least the docs can be changed to reflect the current situation.

### DIFF
--- a/docs/build/reference/integritycheck-require-signature-check.md
+++ b/docs/build/reference/integritycheck-require-signature-check.md
@@ -7,7 +7,7 @@ ms.date: 04/21/2021
 
 Specifies that the digital signature of the binary image must be checked at load time.
 
-> **`/INTEGRITYCHECK`**[**`:NO`**]
+> **`/INTEGRITYCHECK`**
 
 ## Remarks
 
@@ -25,7 +25,11 @@ Microsoft has new signing guidance for DLL and executable files linked by using 
 
 1. Select the **Configuration Properties** > **Linker** > **Command Line** property page.
 
-1. In **Additional Options**, enter *`/INTEGRITYCHECK`* or *`/INTEGRITYCHECK:NO`*. Choose **OK** to save your changes.
+1. To build an image which requires digital signature verification to be loaded, add *`/INTEGRITYCHECK`* to the **Additional Options** command line.
+
+1. To disable the feature, remove the *`/INTEGRITYCHECK`* option.
+
+1. Choose **OK** to save your changes.
 
 ## See also
 
--- a/docs/build/reference/integritycheck-require-signature-check.md
+++ b/docs/build/reference/integritycheck-require-signature-check.md
@@ -25,9 +25,7 @@ Microsoft has new signing guidance for DLL and executable files linked by using 
 
 1. Select the **Configuration Properties** > **Linker** > **Command Line** property page.
 
-1. To build an image which requires digital signature verification to be loaded, add *`/INTEGRITYCHECK`* to the **Additional Options** command line.
-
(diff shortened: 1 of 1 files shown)


glenn-slayden:
c.f. oops... https://github.com/glenn-slayden/cpp-docs/pull/1

Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
Thanks, @glenn-slayden.  I'll take this up with a dev on the linker team and see what direction they'd like to go.

TylerMSFT:
The dev who works on the linker went through the source code history, and sees no evidence of it ever supporting the :NO portion. So that's a strange doc artifact, that your update fixes. Thank you!

docs/build/reference/integritycheck-require-signature-check.md
```diff
@@ -25,7 +25,11 @@ Microsoft has new signing guidance for DLL and executable files linked by using
 
 1. Select the **Configuration Properties** > **Linker** > **Command Line** property page.
 
-1. In **Additional Options**, enter *`/INTEGRITYCHECK`* or *`/INTEGRITYCHECK:NO`*. Choose **OK** to save your changes.
+1. To build an image which requires digital signature verification to be loaded, add *`/INTEGRITYCHECK`* to the **Additional Options** command line.
+
+1. To disable the feature, remove the *`/INTEGRITYCHECK`* option.
```
TylerMSFT, line 30: This might be slightly clearer if we say "This feature is off unless turned on with `/INTEGRITYCHECK`."

TylerMSFT (changes requested):
See what you think of the small suggestion I made. What I'm trying to avoid is people thinking the feature is somehow on unless they do something.

docs/build/reference/integritycheck-require-signature-check.md
TylerMSFT: I updated this. If you have something you'd like better, let me know. In the meantime I'll get your changes through.

TylerMSFT (approved):

TylerMSFT:
#sign-off

glenn-slayden:
@TylerMSFT  Sorry for the late remark after this is closed, but my original text was also not the best:

" &bull; To build an image which requires digital signature verification **_in order_** to be loaded,..."

...would be slightly less ambiguous. It would avoid the incorrect/nonsensical reading where it is somehow the "verification" (as opposed to, correctly, the _image_) which must be "loaded".

TylerMSFT:
No worries, @glenn-slayden. I don't actually that sentence the way you do, but since I'm in this folder today, anyway, I tweaked it like so:
1. To create a digitally-signed image, include `/INTEGRITYCHECK` in the **Additional Options** command line. A digitally-signed image must pass a verification check before it is loaded. This feature is disabled by default.

glenn-slayden:
Super, thanks.

PR #4741, opened 2023-09-30
Update C2345

Rageking8:
`__declspec(align(0))` no longer generates `C2345`, added a high power of 2 to trigger the error.

### DIFF
--- a/docs/error-messages/compiler-errors-1/compiler-error-c2345.md
+++ b/docs/error-messages/compiler-errors-1/compiler-error-c2345.md
@@ -12,11 +12,12 @@ align(value) : illegal alignment value
 
 You passed a value to the [align](../../cpp/align-cpp.md) keyword that is outside the allowable range.
 
-The following code generates C2345
+The following sample generates C2345:
 
 ```cpp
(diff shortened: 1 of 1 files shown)


TylerMSFT (changes requested):
Could you remove the align(0) example altogether? It's a strange beast because it's not a valid alignment (it's ignored). I tried VS 2022, 2019, 2017, and 2015 and none of them generate the C2345 compiler error. I wonder if that was an error by the original writer. At any rate, I like the other examples but would like to remove align(0) because it muddies the waters. 

The spec, [N4964](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4964.pdf) 9.12.2 Alignment specifier [dcl.align]/p4 says:

“The alignment requirement of an entity is the strictest nonzero alignment specified by its alignment-specifiers, if any; otherwise, the alignment-specifiers have no effect.”

So let's not use that (align(0)) example.

TylerMSFT (approved):
Thank you!

TylerMSFT:
#sign-off

PR #4753, opened 2023-10-06
Expand `in-place` tags page

Rageking8:
Not sure if this is the best way since the file name is quite long.


Jak-MS:
@TylerMSFT 
- Can you review this PR?

Note: new articles must be submitted via the private repo so if you accept this, you'll need to move the commits (or we can help). Also, an article deletion requires an entry in the redirection.json (again in the private repo so we can verify that it works)

#label:"aq-pr-triaged"

docs/standard-library/in-place-t-struct-in-place-type-t-struct-in-place-index-t-struct.md
```diff
@@ -0,0 +1,36 @@
+---
+description: "Learn more about: in_place_t Struct, in_place_type_t Struct, in_place_index_t Struct"
+title: "in_place_t Struct, in_place_type_t Struct, in_place_index_t Struct"
+ms.date: "11/04/2016"
+f1_keywords: ["utility<in_place_t>", "utility/std::in_place_t", "utility<in_place_type_t>", "utility/std::in_place_type_t", "utility<in_place_index_t>", "utility/std::in_place_index_t"]
+helpviewer_keywords: ["utility<in_place_t> struct"]
+---
+# in_place_t Struct, in_place_type_t Struct, in_place_index_t Struct
```
TylerMSFT, line 8: Please code escape. Otherwise, auot machine translation may have modify these

TylerMSFT (changes requested):
Hi,
You've been busy :-)
Regarding the PR, we can't delete files. Our policy is not to create broken links if people have bookmarked the topic, etc. If the file name changes, we have to go through a redirection process, which is time consuming.

docs/standard-library/utility.md
```diff
@@ -75,7 +75,7 @@ Pairs are widely used in the C++ Standard Library. They are required both as the
 |-|-|
 |[from_chars_result](../standard-library/from-chars-result-structure.md)|A struct used for `from_chars`.|
 |[identity](../standard-library/identity-structure.md)|A struct that provides a type definition as the template parameter.|
-|[in_place_t](../standard-library/in-place-t-struct.md)|Also includes structs `in_place_type_t` and `in_place_index_t`.|
+|[in_place_t, in_place_type_t, in_place_index_t](../standard-library/in-place-t-struct-in-place-type-t-struct-in-place-index-t-struct.md)|Types used to control in-place construction.|
```
TylerMSFT, line 78: please code escape. I know that this file hasn't been doing that. Many of these were written before we realized what was going on with machine translation, so when we touch a file we code escape functions, types, etc.

docs/standard-library/in-place-t-struct-in-place-type-t-struct-in-place-index-t-struct.md
```diff
@@ -0,0 +1,36 @@
...
+
+The empty structure type `in_place_t` can be passed into constructors of `expected` class, [`optional` class](optional-class.md) or [`single_view` class](single-view-class.md) for in-place construction of the contained type. The empty structure type `in_place_type_t` can be passed into constructors of [`any` class](any-class.md) or [`variant` class](variant-class.md) to specify the type of the object. The empty structure type `in_place_index_t` can be passed into constructors of [`variant` class](variant-class.md) to specify the index of the object.
```
TylerMSFT, line 10: Weren't these introduced in C++17? We should tag them as such and indicate that they should use at least **`/std:c++17`** to compile.

TylerMSFT, line 10: This says where we can pass them, but not why. 
The [ISO spec](https://open-std.org/JTC1/SC22/WG21/docs/papers/2016/n4606.pdf) in section 20.2.7 has some explanation that would be good to make readable and add to the remarks.

TylerMSFT (changes requested):
Hi, thank you for adding to the docs.
There's a deleted file here, which we can't do unless we are willing to go through updating the redirect for the file. I don't think you want to do that. 
We are missing the ## See also section. 
Left other notes in the file.

Rageking8:
@TylerMSFT Would having a separate page for each of the 3 types (`in_place_t`, `in_place_type_t` and `in_place_index_t`) be viable?

TylerMSFT:
> @TylerMSFT Would having a separate page for each of the 3 types (`in_place_t`, `in_place_type_t` and `in_place_index_t`) be viable?

Hi @Rageking8: it's viable, it's just a question of which tradeoff(s) you want to accept. Writing an article is only the first part, and sometimes the smallest part, of the work that an article will require over its lifetime. There's maintenance, metrics, portfolio weeding, etc.  Having lots of tiny files makes that harder, adds to the table of contents which might then make it harder to navigate. When they are related, it makes the reader do more work to pull together the overall picture because they have to click around to different articles. It's a subjective call in the end. I don't see a compelling reason to make them separate topics in this case, but there may be another point of view worth considering. Why would you like to break them up?

TylerMSFT (commented):
@Rageking8 , thank you for providing the base for documenting these types.
The file name change is a problem because it would mean creating a redirect so that the broken link to the old filename redirects to the new. But we don't need a new filename. I also wanted to add an example and more color around what these types are for. 
I'm closing this PR, but your work lives on in [PR #5697](https://github.com/MicrosoftDocs/cpp-docs-pr/pull/5697/files) which I merged today.
Thanks again for your doc contributions. They are most appreciated.

PR #4826, opened 2023-11-22
Small nits for destructors cpp

Rageking8:
Make mention of `new`/`new[]` and `delete`/`delete[]` operators more precise. The empty braces are added to the class list example to prevent invalid syntax highlighting issues. Some other miscellaneous nits added.

### DIFF
--- a/docs/cpp/destructors-cpp.md
+++ b/docs/cpp/destructors-cpp.md
@@ -1,14 +1,14 @@
 ---
 description: "Learn more about: Destructors (C++)"
 title: "Destructors (C++)"
-ms.date: "07/20/2019"
+ms.date: "11/29/2023"
 helpviewer_keywords: ["objects [C++], destroying", "destructors, C++"]
 ---
 # Destructors (C++)
 
-A destructor is a member function that is invoked automatically when the object goes out of scope or is explicitly destroyed by a call to **`delete`** or **`delete[]`**. A destructor has the same name as the class, preceded by a tilde (`~`). For example, the destructor for class `String` is declared: `~String()`.
+A destructor is a member function that is invoked automatically when the object goes out of scope or is explicitly destroyed by a call to **`delete`** or **`delete[]`**. A destructor has the same name as the class and is preceded by a tilde (`~`). For example, the destructor for class `String` is declared: `~String()`.
 
-If you don't define a destructor, the compiler provides a default one; for many classes this is sufficient. You only need to define a custom destructor when the class stores handles to system resources that need to be released, or pointers that own the memory they point to.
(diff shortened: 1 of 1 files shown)


Jak-MS:
@TylerMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

docs/cpp/destructors-cpp.md
```diff
@@ -71,7 +71,7 @@ Destructors are called when one of the following events occurs:
 
 - A local (automatic) object with block scope goes out of scope.
 
-- An object allocated using the **`new`** operator is explicitly deallocated using **`delete`**.
+- An object allocated using the **`new`** or **`new[]`** operator is explicitly deallocated using **`delete`** or **`delete[]`**.
```
TylerMSFT, line 74: I see the parallelism here, but let's make it more explicit since you shouldn't use delete to deallocate memory allocated with new[]. Something like "An object allocated using **`new`** should be deallocated with **`delete`**. An object allocated using **`new[]`** should be deallocated with **`delete[]`**

docs/cpp/destructors-cpp.md
```diff
@@ -124,7 +124,7 @@ int main() {
    delete b2;
 }
 
-Output: A3 dtor
+/* Output: A3 dtor
```
TylerMSFT, line 127: The thing to do here is break this out of the code block, and add a section immediately after that starts:
Three ticks: ```output
A3 dtor
...
```
I've seen the pattern you are trying here in some of our docs. That's an old pattern and I remove it when I come across it, so don't want to introduce it afresh in new docs.

TylerMSFT (changes requested):

docs/cpp/destructors-cpp.md
Rageking8: "\`\`\`Output" with a capital "O" has over 1k matches, but "\`\`\`output" with a lowercase "o" only have 60 something hits. Is any one of the above preferred, or it does not matter?

docs/cpp/destructors-cpp.md
Rageking8: I can't think of a way to make it flow well if we keep to 1 bullet point, hence, I split it into 2. I kept the phrasing of the first part of each bullet point, since it is in the `Destructors are called when one of the following events occurs:` part.

docs/cpp/destructors-cpp.md
TylerMSFT: It doesn't matter because we are relying on the tag for the code block type to do the rendering for us.

docs/cpp/destructors-cpp.md
TylerMSFT: I took a whack at it.

TylerMSFT (approved):
@rageking8, hey, thank you for all the contributions you have made to the C++ docs! I really appreciate the help and hope you have a great holiday season.

TylerMSFT:
#sign-off

PR #4964, opened 2024-03-01
Improve C2548

Rageking8:
Gave the function argument names so as to not confuse people unfamiliar with the `int = 1` syntax, and it is needed for the example explanation. The example explanation is slightly verbose to prevent it from being ambiguous or imprecise, since the default argument requirement is technically the "union" of all preceding declarations within the same scope. Happy to receive feedback on a terser method.


Jak-MS:
@TylerMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

TylerMSFT (approved):
I agree that your changes improve the sample.
All I did was make the description you provided faster to parse.

TylerMSFT:
#sign-off

PR #4975, opened 2024-03-06
Rectify example in C4178

Rageking8:
Rectified some errors in the previous example, namely:
- The usage of an uninitialized variable `i`, which is UB
- The comments on the maximum value being for `unsigned int` but the type of `i` is `int`

The warning is also not emitted with just `/W1` and the default settings of a fresh latest VS2022 C++ project. For some reason, the `/permissive` flag must be set for the warning to be shown. WIth the default of `/permissive-`, the warning is not emitted even with `/Wall` and/or `/w14178`. Not sure if the root cause is some other more granular flag being implicitly set by the permissive flag.

### DIFF
--- a/docs/error-messages/compiler-warnings/compiler-warning-level-1-c4178.md
+++ b/docs/error-messages/compiler-warnings/compiler-warning-level-1-c4178.md
@@ -1,10 +1,9 @@
 ---
 description: "Learn more about: Compiler Warning (level 1) C4178"
 title: "Compiler Warning (level 1) C4178"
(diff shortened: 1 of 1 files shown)


Jak-MS:
@TylerMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

TylerMSFT:
Looking into why /permissive is required.

TylerMSFT (approved):
Thank you. You are correct - the compiler requires /permissive to catch this situation.

PR #4977, opened 2024-03-08
Add example for C2526

Rageking8:
A templated class `A` is used as an un-templated one does not seem to trigger the error.


TylerMSFT (approved):
Much thanks!

TylerMSFT:
#sign-off

PR #4978, opened 2024-03-08
Update C2537

Rageking8:
The second possible cause is removed as it emits [C2733](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2733) instead of this error. In addition, the previous wording implies that "C" is the only supported linkage specifier, which is incorrect.


TylerMSFT (approved):
Thank you!

TylerMSFT:
#sign-off

PR #4981, opened 2024-03-11
Improve C2751

Rageking8:
Fix a few issues:
- Missing `/c` flag
- Adding definitions to the `std` namespace is UB
- Superfluous [C2955](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2955) (use of class template requires template argument list) error

Overall, the example is now slightly simpler.


TylerMSFT (approved):
Thank you for the improvement.

TylerMSFT:
#sign-off

PR #4984, opened 2024-03-13
Tighten wording in mutable data members

Rageking8:
Add missing "non-reference" constraint, since `mutable` is not allowed on reference data members. The empty `main` function is removed as it does not serve any purpose.

### DIFF
--- a/docs/cpp/mutable-data-members-cpp.md
+++ b/docs/cpp/mutable-data-members-cpp.md
@@ -1,14 +1,13 @@
 ---
 description: "Learn more about: Mutable Data Members (C++)"
 title: "Mutable Data Members (C++)"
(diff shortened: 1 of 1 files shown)


Rageking8:
It is already 14 march in my timezone.

TylerMSFT:
> It is already 14 march in my timezone.

The system only cares about its timezone. We can either wait a day or back the date up 1 day.

Rageking8:
I don't mind waiting till tomorrow. Will take note of the date in the future.

Jak-MS:
#label:"aq-pr-triaged"

TylerMSFT (approved):
Thank you.
#signoff

TylerMSFT:
#sign-off

PR #4985, opened 2024-03-13
Update ASAN documentation

branh:
Not all the ASAN code examples actually worked. Some of them require /Od.

This PR updates the examples to be more robust in the face of optimizations, and clarifies the examples where /Od is required and why. There is a corresponding internal PR to make sure we test that these examples still function every time we make changes.

### DIFF
--- a/docs/sanitizers/error-memcpy-param-overlap.md
+++ b/docs/sanitizers/error-memcpy-param-overlap.md
@@ -35,9 +35,11 @@ int main(int argc, char **argv) {
 To build and test this example, run these commands in a Visual Studio 2019 version 16.9 or later [developer command prompt](../build/building-on-the-command-line.md#developer_command_prompt_shortcuts):
 
 ```cmd
-cl example1.cpp /fsanitize=address /Zi
+cl example1.cpp /fsanitize=address /Zi /Oi
 devenv /debugexe example1.exe
 ```
+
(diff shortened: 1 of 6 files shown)


TylerMSFT (approved):
This is great--thank you.

docs/sanitizers/error-memcpy-param-overlap.md
```diff
@@ -35,10 +35,12 @@ int main(int argc, char **argv) {
 To build and test this example, run these commands in a Visual Studio 2019 version 16.9 or later [developer command prompt](../build/building-on-the-command-line.md#developer_command_prompt_shortcuts):
 
 ```cmd
-cl example1.cpp /fsanitize=address /Zi
+cl example1.cpp /fsanitize=address /Zi /Oi
 devenv /debugexe example1.exe
 ```
 
+The [/Oi flag](../build/reference/oi-generate-intrinsic-functions.md) tells the compiler to treat memcpy and memmove as intrinsic functions. This is necessary because some versions of the standard library implement memcpy and memmove in the same way. Because ASAN is a dynamic analysis tool, it only detects errors with an observable runtime effect.
```
ShannonLeavitt, line 42:
```suggestion
The [/Oi flag](../build/reference/oi-generate-intrinsic-functions.md) tells the compiler to treat `memcpy` and `memmove` as intrinsic functions. This is necessary because some versions of the standard library implement `memcpy` and `memmove` in the same way. Because ASAN is a dynamic analysis tool, it only detects errors with an observable runtime effect.
```

branh:
#sign-off

TylerMSFT:
#sign-off

PR #4987, opened 2024-03-16
Update C2860

Rageking8:
Updated the error message (changed since VS2022 17.2 (19.32)):
```diff
- 'void' cannot be an argument type, except for '(void)'
+ 'void' cannot be used as a function parameter except for '(void)'
```
Also tweaked the example.

### DIFF
--- a/docs/error-messages/compiler-errors-2/compiler-error-c2860.md
+++ b/docs/error-messages/compiler-errors-2/compiler-error-c2860.md
@@ -1,22 +1,22 @@
 ---
 description: "Learn more about: Compiler Error C2860"
 title: "Compiler Error C2860"
-ms.date: "11/04/2016"
+ms.date: "03/16/2024"
 f1_keywords: ["C2860"]
 helpviewer_keywords: ["C2860"]
-ms.assetid: ccc83553-90ed-4e94-b5e9-38b58ae38e31
 ---
 # Compiler Error C2860
 
-'void' cannot be an argument type, except for '(void)'
+'void' cannot be used as a function parameter except for '(void)'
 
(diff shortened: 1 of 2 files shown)


docs/error-messages/compiler-errors-2/compiler-error-c2860.md
```diff
@@ -1,22 +1,22 @@
...
-ms.date: "11/04/2016"
+ms.date: "03/16/2024"
 f1_keywords: ["C2860"]
 helpviewer_keywords: ["C2860"]
-ms.assetid: ccc83553-90ed-4e94-b5e9-38b58ae38e31
 ---
 # Compiler Error C2860
 
-'void' cannot be an argument type, except for '(void)'
+'void' cannot be used as a function parameter except for '(void)'
```
Rageking8, line 10: Do we prefer using block quotes for the error message, like:
```diff
- 'void' cannot be used as a function parameter except for '(void)'
+ > 'void' cannot be used as a function parameter except for '(void)'
```

How about the usage of `Remarks` and `Example` subheaders?

TylerMSFT: The 'typical' style for these, for no good reason that I know of other than that's what the rest use, is to avoid block quotes and to not add the ## Example or ## Remarks sections. There are some exceptions if you look around, but they are more the exception than the rule. So we may as well stay with the format most of them are in.

TylerMSFT (approved):

TylerMSFT (approved):
Thank you. Definitely an improvement.

TylerMSFT:
#sign-off

PR #4990, opened 2024-03-18
protected-cpp.md - grammar and consistency

zsotroav:
There is a missing space on line 78 which makes the comment indentation inconsistent and it is annoying me way more than it should...

Also fixed a minor grammar issue along the way.


TylerMSFT (approved):
regarding "There is a missing space on line 78 which makes the comment indentation inconsistent, and it is annoying me way more than it should...", I completely understand :-) Thank you for fixing that for the rest of us as well. Take care.

TylerMSFT:
#sign-off

PR #4993, opened 2024-03-20
Document C2323

Rageking8:
Create dedicated page for error C2323.

### DIFF
--- a/docs/error-messages/compiler-errors-1/compiler-error-c2323.md
+++ b/docs/error-messages/compiler-errors-1/compiler-error-c2323.md
@@ -7,22 +7,22 @@ helpviewer_keywords: ["C2323"]
 ---
 # Compiler Error C2323
 
-'identifier': non-member operator new or delete functions may not be declared static or in a namespace other than the global namespace
+'identifier': non-member operator `new` or `delete` functions may not be declared `static` or in a namespace other than the global namespace.
 
-Overloads for the `new` and `delete` operators must be defined as non-static in the global namespace or as a class member function.
+The `new` and `delete` overload operators must be non-static, defined in the global namespace or as a class members.
 
 The following sample generates C2323:
 
 ```cpp
 // C2323.cpp
 // compile with: /c
-static void* operator new(size_t);   // C2323
-static void operator delete(void*);   // C2323
+static void* operator new(size_t);   // C2323 since static
+static void operator delete(void*);  // C2323 since static
 
 namespace NS
 {
-    void* operator new(size_t);   // C2323
-    void operator delete(void*);   // C2323
+    void* operator new(size_t);   // C2323 since not defined in the global namespace
+    void operator delete(void*);  // C2323 since not defined in the global namespace
 }
 ```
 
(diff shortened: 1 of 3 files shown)


TylerMSFT (approved):
@Rageking8, thank you very much for all your contributions. I really value the improvements you are making!

TylerMSFT:
#sign-off

docs/error-messages/compiler-errors-1/compiler-error-c2323.md
```diff
@@ -0,0 +1,31 @@
...
+description: "Learn more about: Compiler Error C2323"
+ms.date: "03/20/2024"
+f1_keywords: ["C2323"]
+helpviewer_keywords: ["C2323"]
+---
+# Compiler Error C2323
+
+'identifier': non-member operator `new` or `delete` functions may not be declared `static` or in a namespace other than the global namespace.
+
+The `new` and `delete` overload operators must be non-static, defined in the global namespace or as a class members.
```
ShannonLeavitt, line 12:
```suggestion
The `new` and `delete` overload operators must be non-static, defined in the global namespace or as class members.
```
Grammar correction

ShannonLeavitt (approved):

docs/error-messages/compiler-errors-1/compiler-error-c2323.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "Compiler Error C2323"
+description: "Learn more about: Compiler Error C2323"
+ms.date: "03/20/2024"
+f1_keywords: ["C2323"]
+helpviewer_keywords: ["C2323"]
+---
+# Compiler Error C2323
+
+'identifier': non-member operator `new` or `delete` functions may not be declared `static` or in a namespace other than the global namespace.
```
Rageking8, line 10: Should keywords in error messages have backticks or be shown verbatim (raw with no backticks)? Since what I commonly see is the latter.

docs/error-messages/compiler-errors-1/compiler-error-c2323.md
```diff
@@ -0,0 +1,31 @@
...
+
+The `new` and `delete` overload operators must be non-static, defined in the global namespace or as class members.
```
Rageking8, line 12: To be more precise it should be "operator overloads" and not "overload operators".

In addition, with the placement of the comma, one possible misinterpretation of the sentence is that `new`/`delete` operator overloads which are class members must be non-static. That is false as those can be declared `static` and even if the `static` keyword is not explicitly used, it is always a `static` member.

I will address these issues in a follow-up PR.

docs/error-messages/compiler-errors-1/compiler-error-c2323.md
TylerMSFT: The goal is to prevent machine translation. We weren't good about this and got complaints about crazy machine translations for things that shouldn't be translated in the first place. So, we started adding the back ticks.

docs/error-messages/compiler-errors-1/compiler-error-c2323.md
TylerMSFT: That's cool. It was just a bit verbose before.

Rageking8: Maybe we can split it into 2 lines, one for the case where it is in the global namespace (emphasis must be non-static), and the other is for class members (emphasis optional `static` keyword and it being always `static`).

TylerMSFT: Good idea

PR #5010, opened 2024-04-05
Enhance /std documentation with reference to /Zc:__cplusplus

wsfulton:
This change makes it much clearer as to what is required for  C++ standards conformance when using the /std option - Specify Language Standard Version for C++.

Motivation: The /std option implies standards conformance to C++ standards, yet it does not provide the most basic of conformance as it does not predefine the __cplusplus macro as mandated by the C++ standard (in section "Predefined macro names" - [cpp.predefined]). In order to conform to the standard, the /Zc:__cplusplus option is additionally required and this should be very clear in the documentation.

### DIFF
--- a/docs/build/reference/std-specify-language-standard-version.md
+++ b/docs/build/reference/std-specify-language-standard-version.md
@@ -28,6 +28,8 @@ The Microsoft C++ compiler in Visual Studio 2017 and later versions doesn't supp
 
 The **`/std`** option in effect during a C++ compilation can be detected by use of the [`_MSVC_LANG`](../../preprocessor/predefined-macros.md) preprocessor macro. For more information, see [Preprocessor Macros](../../preprocessor/predefined-macros.md).
 
+The [`/Zc:__cplusplus`](zc-cplusplus.md) option must additionally be used for the `__cplusplus` macro to be correctly defined for the appropriate C++ standard. 
+
 **`/std:c++14`**\
 The **`/std:c++14`** option enables C++14 standard-specific features implemented by the MSVC compiler. This option is the default for code compiled as C++. It's available starting in Visual Studio 2015 Update 3.
 
@@ -123,5 +125,6 @@ For more information, see the C Standard library features section of [Microsoft 
 
 ## See also
 
+[`/Zc:__cplusplus[-]`](zc-cplusplus.md)<br/>
 [MSVC compiler options](compiler-options.md)<br/>
 [MSVC compiler command-line syntax](compiler-command-line-syntax.md)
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (approved):
@wsfulton, thank you for making this valuable contribution to the docs. I augmented it a little to provide more context. 
Really appreciate you taking the time to make this update.

TylerMSFT:
#sign-off

Court72:
Thank you, @TylerMSFT!

wsfulton:
Thanks @TylerMSFT, wonderful that this information can be conveyed on the `/std` documentation page and the reformulated wording is much better than the current official documentation which omits this important information.

However, I do feel that the simple one sentence I proposed is perhaps clearer, more relevant and succinct for the `/std` page. The primary simple message on the `/std` page is that one should seriously look at adding in the `/zc:__cplusplus` option and look at the documentation on this page.

I'm not sure repeating the details of the exact values of `__cplusplus` from the `/zc:__cplusplus` page is a good idea, I would leave them on that page and keep a simple link to it. 

The reasons you added for not setting `__cplusplus` according to that mandated by the standard is very useful, but again, getting bogged down with the values in the `/std` page doesn't seem right and is best left on the `/zc:__cplusplus` page.

Obviously it would be the right thing if `/std` did the right thing and was truly standards conforming, but as this is not the case chosen by the MSVC developers, a simple succinct fix for this omission is what a user is interested in and is what the documentation should aim for.

I would suggest this combination:
```
[!IMPORTANT]
The [`/Zc:__cplusplus`](zc-cplusplus.md) option must additionally be used for the `__cplusplus` macro to be correctly defined for the appropriate C++ standard.
Without the addition of `/Zc:__cplusplus`, the value of the `__cplusplus` macro is hard wired to 199711L (c++98) in order to support some existing code that depends on this macro being set to this value.
```

TylerMSFT:
We can't tell people they have to set it and then not tell them to what. I'll keep the table.

wsfulton:
@TylerMSFT, I am sorry, but am a confused by the last comment as users don't need to set the value of `__cplusplus`, the value is set automatically via a combination of `/std` and `/Zc:__cplusplus` options.

If you feel a table is needed on the `/std` page, then please consider copying it from the [zc-cplusplus.md](https://github.com/wsfulton/cpp-docs/blob/c54b897f9594d6ee59d581bd4aa342f69ced2b11/docs/build/reference/zc-cplusplus.md) page, because in comparison:

1. it is technically accurate - does not miss the `L` suffix.
2. it is complete - contains 3 additional rows including `/std:c++latest`.
3. and most importantly, by using 3 columns, is unambiguously and absolutely clear what the `__cplusplus` value is for all combinations of the two options.

TylerMSFT:
I see what you are saying. I will update this in a separate PR. I'll comment on it here when I do so you can see the result.

wsfulton:
Excellent, thanks, I found the commit to main here: 2e67cbc450e766c500e51a8a9bca3c8397e1be48

PR #5015, opened 2024-04-10
Tweak nolock function descriptions

Rageking8:
The wording "locking the thread" seems a little off since it should either be "locking the stream" or some other "item". Omitting the "item" being locked seems to work the best at the expense of being less "specific" (as compared to using the phrase "locking the stream" etc). The large majority of locking for the normal analog of `_nolock` APIs are done via `_lock_file` with a `FILE*` for the stream, with some others being done via `__acrt_lock` with `__acrt_conio_lock` of type `__acrt_lock_id` (both locking functions are implemented using `EnterCriticalSection`) (the term "locking the stream" cannot be used for all `_nolock` APIs). Hence, using the generic term "without locking" for all applicable `_nolock` APIs make things uniform and the idea of "not locking" is decently conveyed.

@TylerMSFT Does this sound like an improvement? Happy to hear some feedback.

Note 1) Some descriptions (e.g. `_fseek_nolock`) are identical to their normal analog, therefore the notion of "not locking" is added.

Note 2) The descriptions of `_getdcwd_nolock` and `_wgetdcwd_nolock` are not modified as they are just macros to the normal analog.

### DIFF
--- a/docs/c-runtime-library/reference/ungetch-ungetwch-ungetch-nolock-ungetwch-nolock.md
+++ b/docs/c-runtime-library/reference/ungetch-ungetwch-ungetch-nolock-ungetwch-nolock.md
@@ -11,7 +11,7 @@ helpviewer_keywords: ["_ungetch function", "ungetwch function", "characters, pus
 ---
 # `_ungetch`, `_ungetwch`, `_ungetch_nolock`, `_ungetwch_nolock`
 
-Pushes back the last character that's read from the console without locking.
+Pushes back the last character that's read from the console.
 
 > [!IMPORTANT]
 > This API cannot be used in applications that execute in the Windows Runtime. For more information, see [CRT functions not supported in Universal Windows Platform apps](../../cppcx/crt-functions-not-supported-in-universal-windows-platform-apps.md).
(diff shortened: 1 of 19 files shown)


docs/c-runtime-library/reference/ungetch-ungetwch-ungetch-nolock-ungetwch-nolock.md
```diff
@@ -11,7 +11,7 @@ helpviewer_keywords: ["_ungetch function", "ungetwch function", "characters, pus
 ---
 # `_ungetch`, `_ungetwch`, `_ungetch_nolock`, `_ungetwch_nolock`
 
-Pushes back the last character that's read from the console without locking.
+Pushes back the last character that's read from the console.
```
Rageking8, line 14: Reverted the addition of "without locking" as this article holds both normal and nolock variants of `ungetch`.

Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (approved):
I agree with you. I think this is an improvement and that being less prescriptive about what exactly is being locked is clearer. Developers will know what locking means in this content.
Thank you for these updates.

TylerMSFT:
#sign-off

PR #4988, opened 2024-03-16
Add links to compiler warnings by compiler version

Rageking8:
Added link to any warnings listed that has its own dedicated page.

PR #5018, opened 2024-04-12
Add some warning entries

Rageking8:
Summary:
- Add some warnings found that are not documented in the warning list page (found while doing #4988).
- Add warning codes that don't have a dedicated page to `f1_keywords`.
- Remove warning codes that have a dedicated page from `f1_keywords`.
- Sync up `f1_keywords` and `helpviewer_keywords`.

Note 1) Normalization of table formatting can be done in a separate PR (maybe to the style of `compiler-warnings-c4800-through-c4999.md`, i.e. with spaces around the table cells).

Note 2) Not sure why article "Compiler warnings C4800 through C5999" has name `compiler-warnings-c4800-through-c4999.md` (this affects the article link as well).


TylerMSFT (approved):
This is good work. Thank you.

Rageking8:
> #sign-off Regarding your question about the filename, I imagine what happened is initially the topic was for the range specified but then expanded. Changing a filename means work to add a redirect so it doesn't break user's bookmarks or anything on the internet that now links to the topic. It can be done, but changing the filename probably didn't rate against the constant flow of new work coming at the team.

One way to fix that issue is to break that article into 2 or more articles. This way no redirect is needed since the article retains it filename, while the excess warnings can be placed in a new article (keeping to the existing 200 warnings/article chunking). Although there will be no link breakage (or redirects), there will still be that "semantic" difference where someone links to that article but are referring to the C5000+ range, which I would argue is rare and not that impactful.

TylerMSFT:
#sign-off
Trying again.

PR #5027, opened 2024-04-24
Recommend `snprintf` over bugprone `_snprinf`

CaseyCarter:
`_snprintf` doesn't terminate the buffer when the format result is overlong, which is just asking for trouble; `snprintf` always terminates. We should recommend the latter to `sprintf` users.


TylerMSFT (approved):
Thanks, Casey!

TylerMSFT:
#sign-off

PR #5053, opened 2024-06-20
Clarify `std::async` policy behavior

Mq-b:
Update async policy description to align with C++ standard and clarify Microsoft implementation

- Modified the description of `launch::async` policy to reflect that the C++ standard states the function behaves as if it invokes the callable object in a new thread.
- Clarified that the Microsoft implementation uses the Windows ThreadPool, which may provide recycled threads, resulting in behavior equivalent to `launch::async|launch::deferred`.
- Added a note about the implications for thread-local variables, recommending against their use with `async`.

This update ensures the documentation accurately reflects both the C++ standard and the specifics of the Microsoft implementation.

References:
- https://eel.is/c++draft/futures.async
- https://en.cppreference.com/w/cpp/thread/launch


Mq-b:
I think the main thing is that C++ doesn't specify:

> The C++ standard states that if policy is launch::async, the function creates a new thread.

Court72:
@TylerMSFT

#label:"aq-pr-triaged"

StephanTLavavej (approved):

TylerMSFT (approved):
Thank you

TylerMSFT:
#sign-off

PR #5056, opened 2024-06-21
Add links for "Just My Code" and "Edit and Continue".

rockchenn:


TylerMSFT:
I fixed the links. We don't allow the locale (en-us) because it breaks localization for people who use another language.
We also prefer repo relative links.
Thank you for adding links here. They are a good addition.

TylerMSFT:
#sign-off

TylerMSFT (approved):

PR #5082, opened 2024-08-16
Fix the description of the _MBC_ILLEGAL return value, especially so in the case `type == _MBC_LEAD`.

WilliamCushing:
Add english descriptions of the conditions that results in illegality, and spell out the value ranges in positive fashion rather than negative fashion based on the values shown in the other rows.

Note that the table does not demonstrate locale-sensitivity or codepage-sensitivity; are the given example values the ones used for codepage 932 (japanese kanji)?  Does locale really effect _mbbtype?


ShannonLeavitt:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (approved):
This is a really nice improvement over the inverted ranges, and the broken range

TylerMSFT:
#sign-off
I'll add addressing your question to my backlog. It'll be a while before I can get to it.

PR #5086, opened 2024-08-23
Fix malformed anchor in headers

Rageking8:
Fix problem where a malformed anchor causes a whole chunk of text to be rendered as a link.

See https://github.com/MicrosoftDocs/visualstudio-docs/pull/10362 for an example of the malformed rendering.

The bulk of the regression was introduced in https://github.com/MicrosoftDocs/cpp-docs/commit/45d31cd4c3bd1ce6343b44c7b3db1452bad1640a.


TylerMSFT (approved):
@rageking8, the format that's in there seems to work fine. I have no objection to the long form so will merge this.

TylerMSFT:
#sign-off

PR #5088, opened 2024-08-23
Update std-specify-language-standard-version.md

CaseyCarter:
* _All_ c++20 Standard Library features have been available with `/std:c++20` since 17.2 and in 16.11 per https://devblogs.microsoft.com/cppblog/msvcs-stl-completes-stdc20/ , leaving 17.0 out as a special case.
* `/std:c++20` has always implied `/permissive-`, which fact we should mention here.
* `/std:c++latest` has also implied `/permissive-` since 16.8, but not in earlier releases.


CaseyCarter:
Support for "`<format>` et al. are available in 16.11 and later except for 17.0: https://www.godbolt.org/z/e774GEWsa

Support for "**`/std:c++20`** implies **`/permissive-`**: https://www.godbolt.org/z/as4rPYj93

Contradiction of "**`/std:c++latest`** implies **`/permissive-`** since 16.11: https://www.godbolt.org/z/GKfs3xsKh. This change was in (16.4, 16.8], I'll run it down before we merge this.

CaseyCarter:
> Contradiction of "**`/std:c++latest`** implies **`/permissive-`** since 16.11: https://www.godbolt.org/z/GKfs3xsKh. This change was in (16.4, 16.8], I'll run it down before we merge this.

Per https://devdiv.visualstudio.com/DevDiv/_git/msvc/pullrequest/268500 this change was made in 16.8 (Actually it was done in the 16.9 timeframe but the change was backported to 16.8 to better support the ongoing implementation of C++20 modules.)

TylerMSFT (approved):
Thanks, Casey!

TylerMSFT:
#sign-off

TylerMSFT:
#hold-off
I'll let you sign off, Casey, since I see a note above about running something down.

TylerMSFT:
#sign-off

CaseyCarter:
Sorry for the messy commit history, I was expecting a squash merge. I'll be more careful in the future.

PR #5090, opened 2024-08-27
_controlfp_s: Describe effects of FPCR changes on ARM

olologin:
We believe this note should be somewhere close to `_controlfp_s` documentation.
More details:
https://stackoverflow.com/questions/78917793/windows-on-arm-math-is-not-trapping

### DIFF
--- a/docs/c-runtime-library/reference/controlfp-s.md
+++ b/docs/c-runtime-library/reference/controlfp-s.md
@@ -75,6 +75,9 @@ _controlfp_s(&current_word, _DN_FLUSH, _MCW_DN);
 // and x64 processors with SSE2 support. Ignored on other x86 platforms.
 ```
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
Hi, we are looking into this.

olologin:
Of course it is better to make it possible to enable exception handling, at least optionally. Because it doesn't look like a good idea to disable them for all ARM CPUs without a way to enable them, just because some ARM CPUs don't support it.

But that is not documentation-related.

TylerMSFT:
@olologin , I am closing this but the reason is that the change affects another file as well. I'm working with the product team on this and want to get the changes all in one PR. I'll notify you via this closed PR when that's done so you can see the change.

TylerMSFT:
@olologin, I updated three docs:
[Overview of ARM64 ABI conventions](https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=msvc-170)
[Overview of ARM64EC ABI conventions](https://learn.microsoft.com/en-us/cpp/build/arm64ec-windows-abi-conventions?view=msvc-170)
[_controlfp_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/controlfp-s?view=msvc-170)

[PR#5850](https://github.com/MicrosoftDocs/cpp-docs-pr/pull/5850/files)

PR #5101, opened 2024-09-14
P0859R0 was implemented in VS 2022 17.1

fsb4000:
Could you recheck the change please? @cdacamar told me that the feature is implemented if I understood correctly.
If not then I will change the same line at https://en.cppreference.com/w/cpp/compiler_support/20 to partial.

### DIFF
--- a/docs/overview/visual-cpp-language-conformance.md
+++ b/docs/overview/visual-cpp-language-conformance.md
@@ -87,7 +87,7 @@ For details on conformance improvements, see [C++ conformance improvements in Vi
 | &emsp;[`P1825R0 Merged wording for P0527R1 and P1155R3, more implicit moves`](https://wg21.link/p1825r0) | VS 2019 16.4 <sup>[17](#note_17)</sup> |
 | &emsp;[`P0929R2 Checking for abstract class types`](https://wg21.link/P0929R2) | VS 2019 16.5 <sup>[17](#note_17)</sup> |
 | &emsp;[`P0962R1 Relaxing the range-for loop customization point finding rules`](https://wg21.link/p0962r1) | VS 2019 16.5 <sup>[17](#note_17)</sup> |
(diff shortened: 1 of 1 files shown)


fsb4000:
@CaseyCarter wrote: "It looks like the work item (VSO-977249) has been "In Progress" since 2021. From the discussion, I get the impression that it's mostly complete. Maybe C++14 support (this was a DR) is not all there? It's not clear to me. For cppreference purposes, I think the fact that we don't define the feature-test macro (https://godbolt.org/z/xcbrv1964) decidedly means we should not claim support."

And I changed the cppreference line to partial. But @TylerMSFT , could you recheck the current status of P0859R0 in MSVC?

cdacamar:
> And I changed the cppreference line to partial. But @TylerMSFT , could you recheck the current status of P0859R0 in MSVC?

Could you please revert that?  The compiler implementation has been in place since 2021 and supported by ([8e97889d258fefe25a9357e1852167832e4b6c37](https://devdiv.visualstudio.com/DevDiv/_git/msvc/commit/8e97889d258fefe25a9357e1852167832e4b6c37).  The feature test macro was never part of the original DR: [P0859R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0859r0.html) and appeared to be added editorially later.

Speaking to why the work item is not complete: internally we used these items known as "UserStory" items to track all of the C++20 work, many of them were not marked as completed due to the automated system (Azure DevOps) being incapable of moving UserStory items -> complete, so many of them are still marked as "In Progress" due to them not being user-visible.

fsb4000:
> Could you please revert that?

Sure, I'll do.

Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
My take from the discussion above is that this change to the docs shouldn't go in. Correct?

CaseyCarter (approved):

CaseyCarter:
> My take from the discussion above is that this change to the docs shouldn't go in. Correct?

Yes. There was confusion that I contributed to which Cameron corrected. The change is good.

TylerMSFT:
#sign-off

Court72 (approved):

PR #5114, opened 2024-10-30
Fix step number references in ARM64 exception handling

mstorsjo:
These step numbers were adjusted in
f510c83085d9764b3c7b52221978af266d54e397, but we forgot to update these references here.

### DIFF
--- a/docs/build/arm64-exception-handling.md
+++ b/docs/build/arm64-exception-handling.md
@@ -393,7 +393,7 @@ Step 6: Allocate remaining stack, including local area, `<x29,lr>` pair, and out
 | 6d | (**CR** == 00 \|\| **CR** == 01) &&<br/>`#locsz` <= 4080 | 1 | `sub sp,sp,#locsz` | `alloc_s`/`alloc_m` |
 | 6e | (**CR** == 00 \|\| **CR** == 01) &&<br/>`#locsz` > 4080 | 2 | `sub sp,sp,4080`<br/>`sub sp,sp,#(locsz-4080)` | `alloc_m`<br/>`alloc_s`/`alloc_m` |
 
-\* If **CR** == 01 and **RegI** is an odd number, Step 2 and the last `save_rep` in step 1 are merged into one `save_regp`.
+\* If **CR** == 01 and **RegI** is an odd number, step 3 and the last `save_reg` in step 2 are merged into one `save_regp`.
 
 \*\* If **RegI** == **CR** == 0, and **RegF** != 0, the first `stp` for the floating-point does the predecrement.
 


ShannonLeavitt:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (commented):
Hi, thank you for updating this topic. I didn't write this one.
I don't see another reference to save_rep in the doc. The reference to "the last save_rep" is the only mention. Is that correct?
As a little nit, would you mind changing the case of Step 2 to step 2.

mstorsjo:
> @mstorsjo, thank you for updating this topic. I don't see another reference to save_rep in the doc. The reference to "the last save_rep" is the only mention. Is that correct?

Oh, thanks for catching this - it’s indeed another typo, it’s supposed to be save_reg; I fixed that now too.

> As a little nit, would you mind changing the case of Step 2 to step 2.

Sure, done.

docs/build/arm64-exception-handling.md
```diff
@@ -393,7 +393,7 @@ Step 6: Allocate remaining stack, including local area, `<x29,lr>` pair, and out
 | 6d | (**CR** == 00 \|\| **CR** == 01) &&<br/>`#locsz` <= 4080 | 1 | `sub sp,sp,#locsz` | `alloc_s`/`alloc_m` |
 | 6e | (**CR** == 00 \|\| **CR** == 01) &&<br/>`#locsz` > 4080 | 2 | `sub sp,sp,4080`<br/>`sub sp,sp,#(locsz-4080)` | `alloc_m`<br/>`alloc_s`/`alloc_m` |
 
-\* If **CR** == 01 and **RegI** is an odd number, Step 2 and the last `save_rep` in step 1 are merged into one `save_regp`.
+\* If **CR** == 01 and **RegI** is an odd number, step 3 and the last `save_reg` in step 2 are merged into one `save_regp`.
```
TylerMSFT, line 396: @mstorsjo, thank you for the updates. I'm still a little confused. There is no save_reg in step 2. There's a save_regp.

mstorsjo: Understandable - it's not spelled out in too much detail.

In step 2, we have one `save_regp_x` followed by more `save_regp` - for each pair of registers that are backed up. If the number of registers is odd, the last one is `save_reg`, not `save_regp`, as the last one doesn't have a pair.

And for that case - if the last one is a lone register, when `RegI` is odd, the save of LR from step 3 (which also would have been a `save_reg`) is merged into a paired store of registers into `save_regp`.

TylerMSFT (approved):
Thank you for the fix and the explanation.

TylerMSFT:
#sign-off

PR #5125, opened 2024-11-18
Update `/MD` flag description to reflect modern behavior

EterDelta:
The current documentation for the `/MD` flag states that applications compiled with this option are statically linked to MSVCRT.lib, which resolves external references to MSVCR.DLL.
However, this is outdated and misleading. In modern VS versions, MSVCRT.lib will prefer the UCRT and redirect to the UCRTBASE.DLL and VCRUNTIME.DLL.
You can check this by compiling a simple binary and dumping its dependencies.

This PR updates the `/MD` description to clarify where the "actual working code" is now.

### DIFF
--- a/docs/build/reference/md-mt-ld-use-run-time-library.md
+++ b/docs/build/reference/md-mt-ld-use-run-time-library.md
@@ -22,7 +22,7 @@ Indicates whether a multithreaded module is a DLL and specifies retail or debug 
 
 |Option|Description|
 |------------|-----------------|
-|**/MD**|Causes the application to use the multithread-specific and DLL-specific version of the run-time library. Defines `_MT` and `_DLL` and causes the compiler to place the library name MSVCRT.lib into the .obj file.<br /><br /> Applications compiled with this option are statically linked to MSVCRT.lib. This library provides a layer of code that enables the linker to resolve external references. The actual working code is contained in MSVCR*versionnumber*.DLL, which must be available at run time to applications linked with MSVCRT.lib.|
+|**/MD**|Causes the application to use the multithread-specific and DLL-specific version of the run-time library. Defines `_MT` and `_DLL` and causes the compiler to place the library name MSVCRT.lib into the .obj file.|
 |**/MDd**|Defines `_DEBUG`, `_MT`, and `_DLL` and causes the application to use the debug multithread-specific and DLL-specific version of the run-time library. It also causes the compiler to place the library name MSVCRTD.lib into the .obj file.|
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
@elterdelta, I'm having misgivings about this page in general. Your correction is right, for a certain time period. The names changed in VS 2015. Because people are still also using versions before that, we'd need to also keep the old names and indicate that the new ones come into play in VS 2015 and later.

What I have misgivings about is that we talk about the actual names at all. During VS 2015, the effort with api sets was to create an abstraction between what you linked against and where the code actually lived. Running dumpbin.exe on a simple console exe would likely show you linking against things like api-ms-win-crt-string-l1-1-0.dll. That will ultimately resolve to ucrtbase.dll, which Windows ships starting in Windows 10, I think.
I wonder if this page should get out of the business of listing the library names and instead say 'the multithreaded version of the runtime library'.  Then, the page could link to the article about the C++ redist, which explains how to install all the runtime dependencies.
You've fixed /MD, but I suspect there are similar issues with a couple of the other flags in this article.
I'm not asking you to do the fixup. But I'm wondering what you'd think about closing this PR and I'll add a work item for myself to do this cleanup.

EterDelta:
Thanks for the context Tyler.

> I wonder if this page should get out of the business of listing the library names and instead say 'the multithreaded version of the runtime library'. Then, the page could link to the article about the C++ redist, which explains how to install all the runtime dependencies.

I don't think linking the C++ redist article is necessary.
The page already links to the [CRT and STL libraries](https://learn.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=msvc-170) article, which already explains you about the UCRT, what was used before, and redistributions:
https://github.com/MicrosoftDocs/cpp-docs/blob/d801435f0b5beeca37b95d29a5f958e75421c746/docs/build/reference/md-mt-ld-use-run-time-library.md?plain=1#L32

> You've fixed /MD, but I suspect there are similar issues with a couple of the other flags in this article.

The other flags descriptions are concise, they don't mention any dependencies. This one being overly descriptive is the problem.

I think the best approach here would be to simply remove this unnecessary listing from the `/MD` description as you said. That way it just looks like the other flags.
We can always add more clarification later if needed.

@TylerMSFT Let me know what you think. I've ammended the commit for simplicity.

TylerMSFT:
Thank you, @EterDelta .  Huge storm event where I live. Power was out a couple days. Now heading into the Thanksgiving holiday so I'll be out for a week.  I'll update this when I get back. I hope your Thanksgiving holiday goes well.

TylerMSFT (approved):
Thank you. I think this is a good approach.

TylerMSFT:
#sign-off

PR #5127, opened 2024-11-25
Fix missing start bracket in `ccomsafearray-class.md`

MathiasVP:
Fixes the missing start parenthesis in the second overload of `operator[]` [here](https://learn.microsoft.com/en-us/cpp/atl/reference/ccomsafearray-class?view=msvc-170#operator_at).


TylerMSFT (approved):
Thank you for fixing that.

TylerMSFT:
#sign-off

PR #5128, opened 2024-11-25
Fix missing parenthesis in `csimplearray-class.md`

MathiasVP:
Similar to https://github.com/MicrosoftDocs/cpp-docs/pull/5127


TylerMSFT (approved):
Thank you

PR #5131, opened 2024-12-02
Update how-to-create-and-use-shared-ptr-instances.md

Nowhere6:
The old description about shared_ptr is wrong. 

If we pass a raw ptr to a callee func, and use it to create a new shared_ptr, it's dangerous because callee will delete the resource where raw ptr pointed to.

Please accept this revision.
image


TylerMSFT (approved):
Thank you for adding clarity around this scenario.

TylerMSFT:
#sign-off

PR #5139, opened 2024-12-09
MASM: Add PROC as another possible type for EXTERN

nir9:
Problem: Current documentation for EXTERN misses an important possible type which is PROC.
This type is important since it enables calling external procedures, for instance, the following
is an example of calling Sleep from kernel32 (assuming the kernel32.lib is specified during linkage):

```
public main
extern Sleep : PROC

.CODE

main:
mov rcx, 1000
call Sleep
ret

END
```


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT (approved):
Thank you.

PR #5143, opened 2024-12-22
Replace all single quotation marks with apostrophes

Rageking8:
Replace both the left and right single quotation mark characters with apostrophes since they are much more prevalent. The quotation marks also cause a compilation error in one of the code examples.


Rageking8:
Wrong branch name.

PR #5144, opened 2024-12-22
Replace all single quotation marks with apostrophes

Rageking8:
Replace both the left and right single quotation mark characters with apostrophes since they are much more prevalent. The quotation marks also cause a compilation error in one of the code examples.


Court72:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT (approved):
Thank you.
I wouldn't sweat the apostrophes, except in the code case where it is clearly wrong. Some of this text was perhaps originally written in Word before being copied into the content authoring system.

TylerMSFT:
#sign-off

PR #5146, opened 2024-12-25
Delete docs/windows/deploying-native-desktop-applications-visual-cpp.md

demetri112:
Fu


ShannonLeavitt:
@demetri112 - I'm afraid articles can't be deleted via the public repo. If you'd like to submit feedback about this article, you can do so at the bottom of the live page:

image

PR #5164, opened 2025-01-29
Update the ABI to support Floating-Point Exceptions.

pmsjt:
Update the ABI to be in concordance, with the introduction of the first officially supported CPU with FP exceptions.

While the original rule was a good intention, it was at odds with the Arm ABI, which is also adopted by Linux and macOS, making code harder to port.

It was also at odds with the Arm Architecture in the sense that if offers a direct way to convey to software when FP exceptions are supported and when they are enabled, and this rule was negating such contract.

### DIFF
--- a/docs/build/arm64-windows-abi-conventions.md
+++ b/docs/build/arm64-windows-abi-conventions.md
@@ -118,7 +118,7 @@ Like AArch32, the AArch64 specification provides three system-controlled "thread
 
 ## Floating-point exceptions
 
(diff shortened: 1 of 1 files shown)


TylerMSFT (approved):
Thank you, Pedro.  When you are ready for this to get merged, add `#sign-off` in a comment.

v-dirichards:
> Thank you, @pmsjt. When you are ready for this to get merged, add `#sign-off` in a comment.

@TylerMSFT I'll merge this PR based on your approval. If this article requires further changes, please open a separate pull request.

PR #5171, opened 2025-02-03
Update check-stack.md

RatinCN:
`#pragma check_stack( )` and `#pragma check_stack` will make stack checking reverts to the behavior specified on the command line, so if `/Gs` presents in command line, stack checking is turn on, otherwise it's turn off.

### DIFF
--- a/docs/preprocessor/check-stack.md
+++ b/docs/preprocessor/check-stack.md
@@ -25,8 +25,8 @@ If you don't give an argument for the **`check_stack`** pragma, stack checking r
 
 | Syntax | Compiled with<br /><br /> `/Gs` option? | Action |
 |--|--|--|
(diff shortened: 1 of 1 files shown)


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
Hi @RatinCN , thank you for taking the time to point this out. I spent time looking at the assembly for different combinations of this flag, and talked to the dev team about it, and I'm going to close this PR because the behavior doesn't line up. In fact, for clarity's sake, one shouldn't use this pragma without arguments.
I've opened PR#5769 to address the behavior of this flag. Thanks for raising this, it was interesting to dig through the actual behavior of the flag.

PR #5176, opened 2025-02-07
Correct return value of `CTaskDialog::OnHelp()`

JoshHarmon:
It returns `S_FALSE`, not `S_OK`, according to `afxtaskdialog.cpp` in 14.42.34433.
image

In testing, returning `S_OK` from an overridden implementation to match the described "default" implementation may result in another window (e.g., a main frame window) also presenting help.

### DIFF
--- a/docs/mfc/reference/ctaskdialog-class.md
+++ b/docs/mfc/reference/ctaskdialog-class.md
@@ -685,7 +685,7 @@ virtual HRESULT OnHelp();
 
 ### Return Value
 
-The default implementation returns S_OK.
+The default implementation returns S_FALSE.
 
 ### Remarks
 


JoshHarmon:
I should note that the fact there is a behavior change based on return value suggests to me that the page for [TDN_HELP](https://learn.microsoft.com/en-us/windows/win32/controls/tdn-help) also is likely to contain bogus claims, such the claim that "The return value is ignored."

TylerMSFT (approved):
@JoshHarmon, thank you for this fix. I verify the same.
I'll notify the owner of the other doc.
As an aside, my verification process included looking at another topic about the CTaskDialog, which I was able to substantially update. So, your good turn had good knock-on effects.

TylerMSFT:
#sign-off

PR #5203, opened 2025-02-24
Tweaks for C6030 warning page

Rageking8:
Summary:
- Remove extra space character in header
- Add links for `[[noreturn]]` and `__declspec(noreturn)`
- Move "Code analysis name" to its own line (every other instance does it that way)
- Remove superfluous trailing empty line in both code snippets
- Format `[[ noreturn ]]` and the function name (prefer snake case for uniformity)

### DIFF
--- a/docs/code-quality/c6030.md
+++ b/docs/code-quality/c6030.md
@@ -6,17 +6,18 @@ f1_keywords: ["C6030", "USE_ATTRIBUTE_NORETURN", "__WARNING_USE_ATTRIBUTE_NORETU
 helpviewer_keywords: ["C6030"]
 ---
 
-# Warning  C6030
+# Warning C6030
 
 > Use attribute [[noreturn]] over __declspec(noreturn) in function '*function-name*'
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


TylerMSFT:
> Just noticed that inline code blocks in link text does not affect the rendered output. For example both of these lines would look the same:
> 
> ```
> [__declspec(noreturn)](https://learn.microsoft.com/en-us/cpp/cpp/noreturn?view=msvc-170)
> [`__declspec(noreturn)`](https://learn.microsoft.com/en-us/cpp/cpp/noreturn?view=msvc-170)
> ```
> 
> However, on GitHub there is a visible difference:
> 
> [__declspec(noreturn)](https://learn.microsoft.com/en-us/cpp/cpp/noreturn?view=msvc-170) [`__declspec(noreturn)`](https://learn.microsoft.com/en-us/cpp/cpp/noreturn?view=msvc-170)
> 
> So should we prefer the former or latter?

The latter (use back-ticks). The reason is that the back ticks prevent automatic machine translation and we wouldn't want translation to happen here. The automatic translator is often smart enough not to try, but rather than play whack-a-mole trying to outsmart it, I default to using the back-ticks.

TylerMSFT (approved):
Thank you

TylerMSFT:
#sign-off

PR #5205, opened 2025-02-24
Update guard-enable-control-flow-guard.md

anporumb:
PR clarifies the options that need to be passed to both the linker and compiler in order to enable/disable CFG.


TylerMSFT:
#sign-off

TylerMSFT (approved):

PR #5211, opened 2025-02-28
Update `CComCoClass` reference page

Rageking8:
Summary:
- Set language of code blocks to `cpp` for syntax highlighting
- Format table and some of the code snippets
- Convert legacy `br` elements to escapes
- Update metadata (`ms.date` is not updated as the changes are mostly superficial)

### DIFF
--- a/docs/atl/reference/ccomcoclass-class.md
+++ b/docs/atl/reference/ccomcoclass-class.md
@@ -1,10 +1,9 @@
 ---
-description: "Learn more about: CComCoClass Class"
 title: "CComCoClass Class"
+description: "Learn more about: CComCoClass Class"
 ms.date: "11/04/2016"
 f1_keywords: ["CComCoClass", "ATLCOM/ATL::CComCoClass", "ATLCOM/ATL::CComCoClass::CreateInstance", "ATLCOM/ATL::CComCoClass::Error", "ATLCOM/ATL::CComCoClass::GetObjectCLSID", "ATLCOM/ATL::CComCoClass::GetObjectDescription"]
 helpviewer_keywords: ["CComCoClass class", "aggregation [C++], aggregation models"]
-ms.assetid: 67cfefa4-8df9-47fa-ad58-2d1a1ae25762
 ---
 # CComCoClass Class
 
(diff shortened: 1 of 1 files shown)


docs/atl/reference/ccomcoclass-class.md
```diff
@@ -1,40 +1,39 @@
...
-|[CComCoClass::Error](#error)|(Static) Returns rich error information to the client.|
-|[CComCoClass::GetObjectCLSID](#getobjectclsid)|(Static) Returns the object's class identifier.|
-|[CComCoClass::GetObjectDescription](#getobjectdescription)|(Static) Override to return the object's description.|
+| Name                                                       | Description                                                             |
+| ---------------------------------------------------------- | ----------------------------------------------------------------------- |
+| [CComCoClass::CreateInstance](#createinstance)             | (Static) Creates an instance of the class and queries for an interface. |
+| [CComCoClass::Error](#error)                               | (Static) Returns rich error information to the client.                  |
+| [CComCoClass::GetObjectCLSID](#getobjectclsid)             | (Static) Returns the object's class identifier.                         |
+| [CComCoClass::GetObjectDescription](#getobjectdescription) | (Static) Override to return the object's description.                   |
```
Rageking8, line 36: Should we format tables as well, to improve readability of the raw markdown source?

ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

docs/atl/reference/ccomcoclass-class.md
TylerMSFT: Good question. I see benefits both ways. I usually don't because it becomes maintenance to keep it that way. As the owner of 10,000 topics I try to reduce maintenance whenever I can.

TylerMSFT (approved):
Thank you!

TylerMSFT:
#sign-off

docs/atl/reference/ccomcoclass-class.md
Rageking8: I would think maintenance becomes a factor for tables that update somewhat frequently. Most of these tables never really change in 5+ years, so it might be worth formatting it once and keeping it as such.

TylerMSFT: It depends on the font in your editor whether the cleaned up version looks cleaned up. And we have editor tools that actually do the opposite and remove all the spaces from tables. I see both versions in our docs. I can see the value both ways. But mostly, I'd prefer not to spend time on PRs that clean up tables because its time I can't really afford for a gain that is nebulous in the scheme of things.
I very much appreciate all the other contributions you make because they add clear customer value and are things I'd have to do myself. You've made ongoing and significant improvements to these docs, for which I thank you.

Rageking8: Noted! Will not format tables in future PRs since there might not be a clear benefit (with potential downsides).

PR #5214, opened 2025-03-01
Flip Compiler Warnings by compiler version table

Rageking8:
Showing the latest version first in the table makes a bit more sense, given that the latest version is more relevant and subsequent sections are also ordered as such. This PR also removes an extra empty line.


TylerMSFT (approved):
Agree! That's a good change.

TylerMSFT:
#sign-off

PR #5215, opened 2025-03-01
Fix off-by-one error in Compiler Warnings by compiler version

Rageking8:
To suppress warnings in the target version and all later versions, the version specified in `/Wv` must be the next older one.


Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
@rageking8, it's Monday; help me out with your thinking here :-)
If a warning is introduced in compiler version 19.43, wouldn't you suppress that warning for 19.43 with /Wv:19.43? I'm wondering why you went a version earlier because we don't want to suppress the warnings in the previous version of the compiler, but 'this' one.

TylerMSFT:
I think you are right. Good catch!

Rageking8:
> I think you are right. Good catch!

Told me some time to work out which was right as well. Might be due to the "whitelist" nature of the flag `/Wv` instead of what I would consider more intuitive at first, "blacklist".

PR #5220, opened 2025-03-05
Remove `19.41` from Compiler Warnings by compiler version

Rageking8:
If there were no new warnings introduced in `19.41`, it should not be included in the table and have its own section, or else it's a contradiction.


ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

TylerMSFT (changes requested):
I'd rather keep this. Otherwise people may think it's an oversight that we missed updating 19.41 and go off looking for that info elsewhere.

PR #5223, opened 2025-03-05
demotion-of-integers.md: Fix example

Alcaro:
-2 isn't 0xFE. `(char)-2` is 0xFE, but IMO adding a type cast would distract from the point of that page; instead, let's switch to integers' default size.


TylerMSFT (approved):
Thank you for this fix. Excellent.

TylerMSFT:
#sign-off

PR #5226, opened 2025-03-07
Remove use of code block language `unstlib`

Rageking8:
Not sure what `unstlib` is, given that there are only 2 occurrences in the whole repository. Hence, this PR replaces `unstlib` with `cpp` and tweaked the `deque` syntax.


TylerMSFT (approved):
I have no idea what unstlib is. I grepped the STL docs and that article seems to be the only one with that weirdness. Thanks for identifying and fixing!

TylerMSFT:
#sign-off

PR #5230, opened 2025-03-08
vsprintf articles update

nikita-leontiev:
1. cd68e32
vswprintf and _vswprintf_l functions return characters count that would be written to buffer when buffer=NULL and count=0. If only count is zero, then memory corruption occurs. "Return value" section updated.
2. 15f1aa3
If buffer=NULL and count=0, functions return characters count that would be written to buffer. If buffer is valid and count/sizeInBytes is zero, then memory corruption occurs. "Return value"/"Remarks" sections updated.
3. 5fa9b33
Functions return characters count that would be written to buffer when buffer=NULL and count=0. If count is zero, functions return -1. "Behavior summary" section updated.
4. 38e139c
If buffer is valid, sizeOfBuffer is not zero and count=0, functions return -1. "Behavior summary" section updated.

Functions behavior was checked on Windows 7 SP1 x64 with Microsoft Visual C++ 2008.


nikita-leontiev:
@microsoft-github-policy-service agree

Court72:
@TylerMSFT

#label:"aq-pr-triaged"

TylerMSFT:
Thank you for this. I have several things in my queue at the moment but have marked this to come back to as soon as I'm able.

TylerMSFT:
Hi @nikita-leontiev, thank you for these updates.

I don't see the corruption issue you refer to. I tested on VS 2022 17.14 and Windows 11.

I did see where vsnprintf_s() we write a fill string to the buffer in Debug builds, and that's by design. I'm going to update the list in the topic that describes this behavior because it's missing from the list.

TylerMSFT:
#sign-off

TylerMSFT (approved):
It took awhile to get to this. Thank you for your patience.

TylerMSFT:
#sign-off

PR #5236, opened 2025-03-13
Update inline-functions-cpp.md

ranjankumar555:
improve readability and concept.


ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

TylerMSFT (changes requested):
I think the original writer was trying to say that the code in the function is substituted. Which 'within' seems to do better than 'in'.

v-dirichards:
> I think the original writer was trying to say that the code in the function is substituted. Which 'within' seems to do better than 'in'.

@TylerMSFT  Would you add a comment to indicate what action to take on this PR?

PR #5241, opened 2025-03-14
Clean up empty row in tables

Rageking8:
Fix 2 instances where the pipe character is misplaced onto a new line, causing the rendered table to have an empty row. The second commit is some basic metadata and `br` element updates.


TylerMSFT (approved):
Empty rows - bad. Thank you.

TylerMSFT:
#sign-off

PR #5245, opened 2025-03-17
Update C2640 error reference

Rageking8:
Summary:
- Update error message to the latest one and add blockquotes
- Add link to the C++ based pointers article
- Replace example with a new one as the old one has multiple issues:
  - No `main` without the `/c` flag
  - Invalid identifier `I`
  - Erroneous assignment to `vr` as it causes an error even if the reference is changed to a pointer (due to incompatible types between `int` and `int __based(vp)*`)
- Edit metadata


TylerMSFT (approved):
Thank you for updating the error message. I like your new example better.

TylerMSFT:
#sign-off

PR #5277, opened 2025-03-29
Fix C2470 example

Rageking8:
Summary:
- Update the error message
- Replace the old example with a new one
  - Old example no longer emits C2470 as `int MyFunc {};` just creates a global variable `MyFunc` initialized to `0`
    - Removing the semicolon to yield the following also does not emit C2470:

      ```cpp
      int MyFunc
      {

      }
      ```
  - Tried many ways but could not get the compiler to emit C2470 due to the "same issue" as the old example (i.e. omitting the parentheses just "converts" it from a function definition to a variable)
  - Settled with a class template member function definition which seems to emit C2470 cleanly (no other errors)
- Edit metadata


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

PR #5278, opened 2025-03-30
Augment C6296 warning reference

Rageking8:
Summary:
- Update warning message (the double spaces are preserved verbatim)
- Improve the generic "might not function as intended." with an emphasis on both *unintentional* and *executes once*
- Augment the possible cause line with the `==` example
- Overhaul example
  - Old example has some flaws (inconsistent formatting and unnecessary splitting of iteration variable)
  - New example incorporates the old one and another `==` instance which emits C6296, together with 2 "normal" for-loops that also execute once showing the "unintentional heuristics"
- Edit metadata

### DIFF
--- a/docs/code-quality/c6296.md
+++ b/docs/code-quality/c6296.md
@@ -1,47 +1,35 @@
 ---
-description: "Learn more about: Warning C6296"
 title: Warning C6296
-ms.date: 11/04/2016
+description: "Learn more about: Warning C6296"
+ms.date: 03/30/2025
 f1_keywords: ["C6296", "LOOP_ONLY_EXECUTED_ONCE", "__WARNING_LOOP_ONLY_EXECUTED_ONCE"]
 helpviewer_keywords: ["C6296"]
-ms.assetid: 226573e0-db18-4c44-8fc6-0bc09d1028bc
 ---
 # Warning C6296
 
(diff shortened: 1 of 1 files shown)


Rageking8:
I am not a fan of the split nature of certain error/warning examples, where the fix is in the second snippet.
Some issues include:
- Potentially huge amounts of duplication depending on the size of the example
- Requires the reader to "manually diff" by scanning both snippets to figure out what changed

Hence, I prefer to provide fixes inline in the same snippet (as seen in this PR). @TylerMSFT any opinions on this matter?

Example screenshot:

image

v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT:
> I am not a fan of the split nature of certain error/warning examples, where the fix is in the second snippet. Some issues include:
> 
> * Potentially huge amounts of duplication depending on the size of the example
> * Requires the reader to "manually diff" by scanning both snippets to figure out what changed
> 
> Hence, I prefer to provide fixes inline in the same snippet (as seen in this PR). @TylerMSFT any opinions on this matter?
> 
> Example screenshot:
> 
> image

I can see where you are coming from. I wouldn't be against combining them when the commented lines to show how to undo the error don't obscure the example.

TylerMSFT:
#sign-off

PR #5313, opened 2025-04-29
Fix invalid range qualification

Rageking8:
The `viewable_range` concept belongs in the `std::ranges` namespace, hence the addition of the missing "s".


TylerMSFT:
#sign-off

PR #5340, opened 2025-05-21
Update asan-known-issues.md

zacklj89:
- Remove Win10 reference, as we're **always** targeting the latest Windows release with ASan
- Add known issues bit about unloading DLLs causing issues

### DIFF
--- a/docs/sanitizers/asan-known-issues.md
+++ b/docs/sanitizers/asan-known-issues.md
@@ -83,16 +83,16 @@ Thread local variables (global variables declared with `__declspec(thread)` or `
 
 If all of the code in a process isn't compiled with `/fsanitize=address`, ASan may not be able to diagnose all memory safety errors. The most common example is when a DLL is compiled with ASan but is loaded into a process that contains code that wasn't compiled with ASan. In this case, ASan attempts to categorize allocations that took place prior to ASan initialization. But if those allocations are reallocated, moved, and so on, ASan tries to own and monitor the lifetime of memory loads and stores.
 
-If all of the DLLs that were compiled with ASan are unloaded from the process before the process ends (as is the case with test drivers such as [TAEF](https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/)), there may be crashes due to dangling references to intercepted functions such as `memcmp`, `memcpy`, `memmove`, and so on.
+If all of the DLLs that were compiled with ASan are unloaded from the process before the process ends (as is the case with test drivers such as [TAEF](/windows-hardware/drivers/taef/))), there may be crashes due to dangling references to intercepted functions such as `memcmp`, `memcpy`, `memmove`, and so on.
 
 Please report any bugs to our [Developer Community](https://aka.ms/feedback/report?space=62).
 
 ## See also
 
-[AddressSanitizer overview](./asan.md)\
-[AddressSanitizer build and language reference](./asan-building.md)\
-[AddressSanitizer runtime reference](./asan-runtime.md)\
-[AddressSanitizer shadow bytes](./asan-shadow-bytes.md)\
-[AddressSanitizer cloud or distributed testing](./asan-offline-crash-dumps.md)\
-[AddressSanitizer debugger integration](./asan-debugger-integration.md)\
-[AddressSanitizer error examples](./asan-error-examples.md)
+[AddressSanitizer overview](asan.md)\
+[AddressSanitizer build and language reference](asan-building.md)\
+[AddressSanitizer runtime reference](asan-runtime.md)\
+[AddressSanitizer shadow bytes](asan-shadow-bytes.md)\
+[AddressSanitizer cloud or distributed testing](asan-offline-crash-dumps.md)\
+[AddressSanitizer debugger integration](asan-debugger-integration.md)\
+[AddressSanitizer error examples](asan-error-examples.md)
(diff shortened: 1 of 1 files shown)


docs/sanitizers/asan-known-issues.md
```diff
@@ -82,6 +79,14 @@ As a workaround, create a *`Directory.Build.props`* file in the root of your pro
 
 Thread local variables (global variables declared with `__declspec(thread)` or `thread_local`) aren't protected by AddressSanitizer. This limitation isn't specific to Windows or Microsoft Visual C++, but is a general limitation.
 
+## Loading and unloading DLLs
```
davidmrdavid, line 82: I'm partial to this alternative title.

```suggestion
## Issues with partially sanitized executables.
```

After all, the issue is not with loading and unloading dlls per-se, but instead with the partial `/fsanitize` across components. What do you think?

docs/sanitizers/asan-known-issues.md
```diff
@@ -82,6 +79,14 @@ As a workaround, create a *`Directory.Build.props`* file in the root of your pro
...
+
+If the entire process is not compiled with `/fsanitize=address`, there may be limitations in ASan's ability to diagnose memory safety errors. The most common example is when a DLL is compiled with ASan and loaded into a process that **was not** compiled with ASan. In this case, ASan will attempt to categorize allocations that took place prior to ASan initialization. If those allocations are transformed in any way (reallocated, moved, etc.), ASan will then attempt to own and monitor the lifetime and memory loads/stores.
```
davidmrdavid, line 84: nit: I think we can be a bit more assertive in the opening sentence:

```suggestion
ASan's analysis is limited when executing alongside binaries that were not complied with `/fsanitize=address`. For example, this would occur if an executable compiled _without_ `/fsanitize=address` loads an ASan-instrumented `.dll`.

In this case, ASan coverage over allocations that took place prior to ASan initialization is best-effort. If those allocations are transformed in any way (reallocated, moved, etc.), ASan will then attempt to own and monitor their lifetime and accesses moving forward.
```

docs/sanitizers/asan-known-issues.md
```diff
@@ -82,6 +79,14 @@ As a workaround, create a *`Directory.Build.props`* file in the root of your pro
...
+
+In the event that all DLLs compiled with ASan leave the process before the end of its execution (like in the event of test drivers such as [TAEF](https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/)), there may dangling references to intercepted functions (`memcmp`, `memcpy`, `memmove`, etc.), and may result in crashes. This is a current limitation with these configurations and unloading DLLs.
```
davidmrdavid, line 86: in this case, I think the `TAEF` reference may be removed, unless we think we'll get more TAEF reports in the future. I don't feel strongly, but given that it's mostly an internal tool, I'm unsure if there's much value in referencing it externally.

davidmrdavid: > This is a current limitation with these configurations and unloading DLLs.

Is this a limitation of TAEF, or our own. I feel like there's little _we_ could do to improve this, or is there?

davidmrdavid (commented):
Thanks for the update. Left some comments, feel free to push back

docs/sanitizers/asan-known-issues.md
v-dirichards: @tylermsft Can you respond to the comment above?

v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

docs/sanitizers/asan-known-issues.md
TylerMSFT: This isn't my article. @zacklj89 is writing it. 
My 2c is I like the suggested change.

zacklj89: One reason I didn't use _sanitized_ in the title is there is no previous mention of "sanitized" here (whereas we define ASan, STL, etc). I wanted to avoid confusion by naming the section as the action that is taken when the side effects occur. I'll let @TylerMSFT have the final call.

TylerMSFT: I think the reader can make the leap, given the context of the article, and that the suggested heading more directly addresses why the reader might want to read that section.

TylerMSFT (changes requested):
Let me know when you've resolved the little issues and we'll get this merged.

docs/sanitizers/asan-known-issues.md
```diff
@@ -26,9 +26,9 @@ These options and functionality are incompatible with [`/fsanitize=address`](../
 
 ## Standard library support
 
-The MSVC standard library (STL) is partially enlightened to understand the AddressSanitizer and provide other checks. For more information, see [container-overflow error](./error-container-overflow.md).
+The MSVC standard library (STL) makes partial use of the AddressSanitizer and provides other code safety checks. For more information, see [container-overflow error](./error-container-overflow.md).
 
-When annotations are disabled or in versions without support, AddressSanitizer exceptions raised in STL code do still identify true bugs. However, they aren't as precise as they could be.
+When annotations are disabled, or in versions of the Standard Library that don't support them, AddressSanitizer exceptions raised in STL code still identify real bugs. However, they are more precise if annoations are enabled and you use a version of the Standard Library that supports them.
```
Rageking8, line 31:
```suggestion
When annotations are disabled, or in versions of the Standard Library that don't support them, AddressSanitizer exceptions raised in STL code still identify real bugs. However, they are more precise if annotations are enabled and you use a version of the Standard Library that supports them.
```

zacklj89:
> Let me know when you've resolved the little issues and we'll get this merged.

@TylerMSFT let me know how this looks!

zacklj89:
#sign-off

TylerMSFT:
#sign-off

PR #5348, opened 2025-05-25
Fix inverted order of inline code and italics

Rageking8:
Fix inverted order of inline code and italics which causes it to be rendered verbatim, for example:

image


v-regandowner:
@TylerMSFT - Can you review the proposed changes? 
	
#label:"aq-pr-triaged"

TylerMSFT:
#sign-off

PR #5349, opened 2025-05-25
Remove stray semicolons after `#include` directive

Rageking8:
Remove 2 instances of stray semicolons after `#include` directive.


TylerMSFT:
#sign-off

PR #5351, opened 2025-05-26
Clean up extraneous backticks

Rageking8:
Remove backticks that are not needed.


TylerMSFT (approved):
I guess the original author was really making sure ;-)

PR #5369, opened 2025-05-30
Tweak C4789 warning reference

Rageking8:
Remove stray backtick, compact unordered list, use colons, and update metadata.


Rageking8:
> I get curious about things like the stray backtick. I wonder if someone didn't realize that their editor had the focus and they tried to type somewhere else and didn't realize they left this behind. But in what other context would you be typing a backtick? These are the mysteries.

Some instances of `CXXXX` error codes are backticked in the docs. It could then be possible the author forgot to add a matching backtick or was somehow interrupted after entering 1 backtick.

PR #5370, opened 2025-05-30
Fix wrong remarks for `forward_list::emplace_front` and add backticks

Rageking8:
Details of the remark fix can be found in the review comment.

### DIFF
--- a/docs/standard-library/forward-list-class.md
+++ b/docs/standard-library/forward-list-class.md
@@ -350,7 +350,7 @@ The element added to the beginning of the forward list.
 
 ### Remarks
 
-This member function inserts an element with the constructor arguments `_ val` at the end of the controlled sequence.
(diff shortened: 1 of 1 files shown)


docs/standard-library/forward-list-class.md
```diff
@@ -345,16 +345,16 @@ template <class Type>
...
 ### Parameters
 
-*val*\
+*`val`*\
 The element added to the beginning of the forward list.
 
 ### Remarks
 
-This member function inserts an element with the constructor arguments `_ val` at the end of the controlled sequence.
+This member function inserts an element with the constructor arguments `val` at the beginning of the controlled sequence.
```
Rageking8, line 353: There is no such argument as `_ val` and `emplace_front` inserts to the beginning and not end.

TylerMSFT: Should I laugh or cry here? Good thing its Friday.

PR #5387, opened 2025-06-04
Improve C2574 error reference

Rageking8:
- Improve the error message to account for both constructors and destructors
- Add "Remarks" and "Example" headings to keep things clean
- Tweak remarks (constructor before destructor seems more canonical) and add links
- Overhaul example
  - Use struct instead, to avoid needing `public:` and prevent issues with defining the members as `private`
  - Remove orthogonal `virtual` to keep the example more focused
  - Define instead of just declaring the constructor to prevent `VCR001` warning
- Update metadata


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT (approved):
That's a better example. Very nice.

TylerMSFT:
#sign-off

PR #5396, opened 2025-06-08
Fix wrong links to "Heterogeneous Lookup in Associative Containers"

Rageking8:
Not sure why so many of these links point to the wrong `sequence_containers` anchor. The `heterogeneous-lookup-in-associative-containers` anchor is added as the default `heterogeneous-lookup-in-associative-containers-c14` one might give wrong connotations about C (although C14 does not actually exist). Kebab case is chosen over snake case since the default one is as such.

### DIFF
--- a/docs/standard-library/multimap-class.md
+++ b/docs/standard-library/multimap-class.md
@@ -30,7 +30,7 @@ The element data type to be stored in the multimap.
 *`Traits`*\
 The type that provides a function object that can compare two element values as sort keys to determine their relative order in the multimap. The binary predicate `less<Key>` is the default value.
 
-In C++14 you can enable heterogeneous lookup by specifying the `std::less<>` or `std::greater<>` predicate that has no type parameters. For more information, see [Heterogeneous Lookup in Associative Containers](../standard-library/stl-containers.md#heterogeneous-lookup-in-associative-containers-c14)
+In C++14 you can enable heterogeneous lookup by specifying the `std::less<>` or `std::greater<>` predicate that has no type parameters. For more information, see [Heterogeneous Lookup in Associative Containers](../standard-library/stl-containers.md#heterogeneous-lookup-in-associative-containers)
 
 *`Allocator`*\
 The type that represents the stored allocator object that encapsulates details about the map's allocation and deallocation of memory. This argument is optional and the default value is `allocator<pair <const Key, Type> >`.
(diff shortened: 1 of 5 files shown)


v-regandowner:
@TylerMSFT - Can you review the proposed changes? 
	
#label:"aq-pr-triaged"

docs/standard-library/multimap-class.md
```diff
@@ -30,7 +30,7 @@ The element data type to be stored in the multimap.
 *`Traits`*\
 The type that provides a function object that can compare two element values as sort keys to determine their relative order in the multimap. The binary predicate `less<Key>` is the default value.
 
-In C++14 you can enable heterogeneous lookup by specifying the `std::less<>` or `std::greater<>` predicate that has no type parameters. For more information, see [Heterogeneous Lookup in Associative Containers](../standard-library/stl-containers.md#heterogeneous-lookup-in-associative-containers-c14)
```
TylerMSFT, line 33: Are you sure about removing the -c14? 
When I look at the link in the browser, its #heterogeneous-lookup-in-associative-containers-c14

TylerMSFT (changes requested):
I have a question about the removal of -c14. Looks like it is necessary. This is what I see when I copy the anchor link in the browser:
#heterogeneous-lookup-in-associative-containers-c14

docs/standard-library/multimap-class.md
Rageking8: Since I added a custom anchor `heterogeneous-lookup-in-associative-containers`, that will be used instead (i.e. without the "-c14"). The "-c14" is present in the browser as we don't have a custom anchor, and the default one is being used. I have cross checked with other instances and verified that doing `#custom-anchor` works. Hope that clears it up.

docs/standard-library/stl-containers.md
```diff
@@ -39,7 +38,7 @@ Both `map` and `set` only allow one instance of a key or element to be inserted
 
 Ordered maps and sets support bi-directional iterators, and their unordered counterparts support forward iterators. For more information, see [Iterators](../standard-library/iterators.md).
 
-### Heterogeneous Lookup in Associative Containers (C++14)
+### <a name="heterogeneous-lookup-in-associative-containers"></a> Heterogeneous Lookup in Associative Containers (C++14)
```
Rageking8, line 41: Here is the aforementioned custom anchor.

docs/standard-library/multimap-class.md
TylerMSFT: I missed your addition of the custom anchor. All good :-)

TylerMSFT:
#sign-off

PR #5406, opened 2025-06-11
Fix duplicate `operator[]` entry in `CSimpleStringT` reference

Rageking8:
Discovered via the duplicate `operator_at` anchor value. There does not seem to be another overload of `operator[]`, hence the extra entry is removed. Both entries are slightly different, but the first one is kept since it has the example. The diff between the first and second entry:

````diff
- ## <a name="operator_at"></a> `CSimpleStringT::operator[]`
+ ## <a name="operator_at"></a> `CSimpleStringT::operator []`

  Call this function to access a single character of the character array.

  ### Syntax

  ```cpp
  XCHAR operator[](int iChar) const;
  ```

- #### Parameters
+ ### Parameters

  *`iChar`*<br/>
  Zero-based index of a character in the string.

  ### Remarks

  The overloaded subscript (**`[]`**) operator returns a single character specified by the zero-based index in *`iChar`*. This operator is a convenient substitute for the [`GetAt`](#getat) member function.

  > [!NOTE]
  > You can use the subscript (**`[]`**) operator to get the value of a character in a `CSimpleStringT`, but you cannot use it to change the value of a character in a `CSimpleStringT`.
-
- ### Example
-
- The following example demonstrates the use of `CSimpleStringT::operator []`.
-
- ```cpp
- CSimpleString s(_T("abc"), pMgr);
- ASSERT(s[1] == _T('b'));
- ```
````

### DIFF
--- a/docs/atl-mfc-shared/reference/csimplestringt-class.md
+++ b/docs/atl-mfc-shared/reference/csimplestringt-class.md
@@ -716,28 +716,6 @@ CSimpleString s(_T("abc"), pMgr);
 ASSERT(s[1] == _T('b'));
 ```
 
-## <a name="operator_at"></a> `CSimpleStringT::operator []`
-
-Call this function to access a single character of the character array.
-
-### Syntax
-
-```cpp
-XCHAR operator[](int iChar) const;
-```
-
-### Parameters
-
-*`iChar`*<br/>
-Zero-based index of a character in the string.
-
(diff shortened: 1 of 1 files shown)


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT:
#sign-off

PR #5430, opened 2025-06-17
Added documentation for new clang-tidy settings

Omotola:
Added information for 3 new clang-tidy properties that allow uses to pass extra arguments to the tool and specify number of processes to run.


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

docs/code-quality/clang-tidy.md
```diff
@@ -94,6 +94,18 @@ By default, Clang-Tidy does not set any checks when enabled. To see the list of
...
+
+## Clang-Tidy Prepend Additional Options
+
+The **Clang-Tidy Prepend Additional Options** property lets you specify compiler arguments that are passed to Clang-Tidy using the `--extra-args-before` command-line option. These arguments are inserted before the default compiler arguments when Clang-Tidy parses your code. Enter arguments as a semi-colon separated list. For example: `-std=c++20;-DMY_DEFINE=1;`.
+
+## Max Number of Processes
+
+The **Max Number of Processes** property lets you specify how many processes Clang-Tidy can use to run analysis in parallel. By default, Clang-Tidy uses a single processor. Set this property to a specific number to limit the number of parallel processes, or set it to `0` to use all available logical processors on your system. Increasing the number of processes can improve analysis speed on multi-core machines.
```
TylerMSFT: I'm unclear how 'uses a single processor' fits in here. The text seems to say that by default Clang-Tidy uses a single processor, yet you can set the property to a specific number to limit the number of parallel processes. Why would you need to limit it if the default is to run on a single processor?
Maybe the problem here is the difference between a physical processor and logical processes. Would it be correct to instead say something like "Clang-Tidy uses a single processor to run an unlimited number of logical processes in parallel. Set this property to limit the number of processes run in parallel. Set it to `0` to use all available logical processors in your system."

Or is the physical processor a red-herring here? If there are multiple processes, why are they run on a single processor?

At any rate, I'm confused by how a physical processor plays into the logical processors that the property seems to be about.

TylerMSFT (changes requested):

docs/code-quality/clang-tidy.md
Omotola: I think the wording around "limit the number of parallel processes" is where the confusion is. Clang-tidy runs serially by default and this new setting is to enable users run it in parallel. I used the word "limit" to note that if users don't want to set to '0' to use all available processors they can limit the number of processes instead. 
What do you think of this wording instead - "By default, Clang-Tidy runs serially. Set this property to enable parallel execution and specify the number of parallel processes. Set it to 0 to use all available logical processors in your system."

TylerMSFT: That sounds great to me.

TylerMSFT (approved):
Looks good. Thank you, @Omotola

TylerMSFT:
#sign-off

PR #5431, opened 2025-06-17
Fix documentation of __STDC__

puetzk:
getting `__STDC__=1` requires setting `/Za` or `/Zc:__STDC__`. It doesn't seem to matter whether/what `/std` is set to.

As best I can tell (experimenting locally with Visual Studio 2022 17.14, and on compiler explorer back to 17.2), the changes made in c1398171f4ecdf74585fba486debf2796ca9d4d5 do not seem to be accurate. setting `/std:c11` or `/std:c17` do not result in having `#define __STDC__ 1`. See https://godbolt.org/z/65YvxPYM3

This agrees with the documentation of https://learn.microsoft.com/en-us/cpp/build/reference/zc-stdc?view=msvc-170, which says

> This option is a source breaking change. Due to the behavior of the UCRT, which doesn't expose POSIX functions when __STDC__ is 1, it isn't possible to define this macro for C by default without introducing breaking changes to the stable language versions.

So I suspect having `/std:c11` set `__STDC__=1` may have originally been the original intent (and maybe it was true in a preview version?) but it doesn't seem to behave that way in the 17.2 or 17.14 release...


Rageking8 (approved):
Correct as per [\_\_STDC__ not defined with std:c11](https://developercommunity.visualstudio.com/t/__STDC__-not-defined-with-std:c11/10498060#T-N10498384).

v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT (approved):
This is correct. Thank you.

TylerMSFT:
#sign-off

TylerMSFT:
#sign-off

PR #5474, opened 2025-06-29
Remove outdated example and update C2998 error reference

Rageking8:
The old example is presumably written before variable template got introduced in C++14, which made that definition well formed. I tried and could not get the compiler to emit C2998, hence the example should be removed for now.

### DIFF
--- a/docs/error-messages/compiler-errors-2/compiler-error-c2998.md
+++ b/docs/error-messages/compiler-errors-2/compiler-error-c2998.md
@@ -9,4 +9,6 @@ helpviewer_keywords: ["C2998"]
 
 > '*identifier*': cannot be a template definition
 
+## Remarks
+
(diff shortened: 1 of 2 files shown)


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

v-dirichards:
@tylermsft  Could you review this proposed update to your article and enter `#sign-off` in a comment if it's ready to merge?
	
Thanks!

TylerMSFT (approved):
I tried several examples and the compiler is better at calling out the specific thing that is wrong rather than giving this more general error. I was hoping template <class T> enum MyEnum { Value }; // C2998 would do it, but no. Agree with pulling the example.

TylerMSFT:
#sign-off

PR #5519, opened 2025-07-09
Add missing bracket to modules-cpp.md

JohnVeness:
It was accidentally removed in Commit 1b29af4.


v-dirichards (approved):

JohnVeness:
Actually, you know what, on another reading I think I've still not corrected this properly! Although I've put it back to how it was before 1b29af4, I think it was wrong then too. I think it should be:

A module is a set of source code files that are compiled independently of the source files (or more precisely, the translation units) that import them.

Another PR incoming!

PR #5537, opened 2025-07-14
Standardize paragraph about ASCII, SBCS, and MBCS

Rageking8:
Standardize the 2 slightly differing paragraphs about ASCII, SBCS, and MBCS. The motivation was the phrasing in [`docs/c-runtime-library/single-byte-and-multibyte-character-sets.md`](https://github.com/MicrosoftDocs/cpp-docs/blob/df4242d2a81e97ed4da3e0f11672f8336c09da26/docs/c-runtime-library/single-byte-and-multibyte-character-sets.md):

> ... Japanese Kanji, include many more characters *than can* be represented in a single-byte coding scheme, and so ...

Although technically correct, the "than can" is prone to being misread as "that can" (I fell for it...), which in this case flips the meaning.


ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

TylerMSFT:
#sign-off

PR #5573, opened 2025-07-23
Improve C2469 error reference

Rageking8:
- Change error message to one with explicit reference to `new` and `void` as it seems newer (commit https://github.com/MicrosoftDocs/cpp-docs/commit/062031c7e1454df22636fe97788b729f8a24a0e8 added this variant to the index page, while the older one is there since the first commit in 2016) and the error does not seem to apply to anything other than operator `new` with type `void`.
- Add "Remarks" heading and tweak its contents to reflect the new error message.
- Overhaul the existing example and add a new one based on the user's intent (i.e. they specified the wrong allocation type versus they made an invalid attempt to allocate untyped memory (see [c++ - Using new on void pointer - Stack Overflow](https://stackoverflow.com/questions/14111900/using-new-on-void-pointer))).
- Update metadata

### DIFF
--- a/docs/error-messages/compiler-errors-1/compiler-error-c2469.md
+++ b/docs/error-messages/compiler-errors-1/compiler-error-c2469.md
@@ -1,7 +1,7 @@
 ---
 title: "Compiler Error C2469"
 description: "Learn more about: Compiler Error C2469"
-ms.date: 07/23/2025
+ms.date: 2/27/2026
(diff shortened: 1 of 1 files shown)


ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

TylerMSFT:
#sign-off

TylerMSFT:
#sign-off

PR #5574, opened 2025-07-23
Tweak "Raw pointers (C++)" topic

Rageking8:
Fix a bunch of mistakes and perform some improvements. See review comments for more information.

### DIFF
--- a/docs/cpp/raw-pointers.md
+++ b/docs/cpp/raw-pointers.md
@@ -78,7 +78,7 @@ void func_B(MyClass mc)
     // This statement modifies only the local copy of mc.
     mc.num = 21;
     std::cout << "Local copy of mc:";
-    mc.print(); // "Erika, 21"
+    mc.print(); // "Erika:21"
 }
 
 int main()
@@ -99,25 +99,25 @@ int main()
     MyClass* pcopy = &mc;
 
     // Use the -> operator to access the object's public members
-    pmc->print(); // "Nick, 108"
+    pmc->print(); // "Nick:108"
 
     // Copy the pointer. Now pmc and pmc2 point to same object!
     MyClass* pmc2 = pmc;
 
     // Use copied pointer to modify the original object
     pmc2->name = "Erika";
-    pmc->print(); // "Erika, 108"
-    pmc2->print(); // "Erika, 108"
+    pmc->print(); // "Erika:108"
+    pmc2->print(); // "Erika:108"
 
     // Pass the pointer to a function.
     func_A(pmc);
-    pmc->print(); // "Erika, 3"
-    pmc2->print(); // "Erika, 3"
+    pmc->print(); // "Erika:3"
+    pmc2->print(); // "Erika:3"
 
     // Dereference the pointer and pass a copy
     // of the pointed-to object to a function
     func_B(*pmc);
-    pmc->print(); // "Erika, 3" (original not modified by function)
+    pmc->print(); // "Erika:3" (original not modified by function)
 
     delete pmc; // don't forget to give memory back to operating system!
     // delete pmc2; //crash! memory location was already deleted


docs/cpp/raw-pointers.md
```diff
@@ -19,7 +19,7 @@ A pointer can also be *dereferenced* to retrieve the value of the object that it
...
+A pointer can point to a typed object or to **`void`**. When a program allocates an object on the [heap](https://wikipedia.org/wiki/Heap) in memory, it receives the address of that object in the form of a pointer. Such pointers are called *owning pointers*. An owning pointer (or a copy of it) must be used to explicitly free the heap-allocated object when it's no longer needed. Failure to free the memory results in a *memory leak*, and renders that memory location unavailable to any other program on the machine. Memory allocated with **`new`** must be freed using **`delete`**, and memory allocated with **`new[]`** must be freed using **`delete[]`**. For more information, see [`new` and `delete` operators](new-and-delete-operators.md).
```
Rageking8, line 22: > Memory allocated using **`new`** must be freed by using **`delete`** (or **`delete[]`**).

This implies `delete[]` can be used in place of `delete`, which is wrong. Hence, this sentence is extended to make  it explicit that `new` must be matched with `delete`, and `new[]` must be matched with `delete[]`.

docs/cpp/raw-pointers.md
```diff
@@ -78,7 +78,7 @@ void func_B(MyClass mc)
     // This statement modifies only the local copy of mc.
     mc.num = 21;
     std::cout << "Local copy of mc:";
-    mc.print(); // "Erika, 21"
+    mc.print(); // "Erika:21"
```
Rageking8, line 81: The output comments used the wrong delimiter (", " instead of the correct ":"). All subsequent occurrences are fixed.

docs/cpp/raw-pointers.md
```diff
@@ -99,28 +99,28 @@ int main()
...
     // Dereference the pointer and pass a copy
     // of the pointed-to object to a function
     func_B(*pmc);
-    pmc->print(); // "Erika, 3" (original not modified by function)
+    pmc->print(); // "Erika:3" (original not modified by function)
 
-    delete(pmc); // don't forget to give memory back to operating system!
-   // delete(pmc2); //crash! memory location was already deleted
+    delete pmc; // don't forget to give memory back to operating system!
+    // delete pmc2; //crash! memory location was already deleted
```
Rageking8, line 123: Prefer the more idiomatic `delete expr;` over `delete(expr);`, even though the latter technically works. As an aside, found this interesting comment at [c++ - Is there any difference between delete x and delete(x)? - Stack Overflow](https://stackoverflow.com/questions/29043061/is-there-any-difference-between-delete-x-and-deletex#comment46326145_29043061).

docs/cpp/raw-pointers.md
```diff
@@ -156,7 +156,11 @@ int main()
 }
 ```
 
-Certain arithmetic operations can be used on non-`const` pointers to make them point to another memory location. Pointers are incremented and decremented using the **`++`**, **`+=`**, **`-=`** and **`--`** operators. This technique can be used in arrays and is especially useful in buffers of untyped data. A `void*` gets incremented by the size of a **`char`** (1 byte). A typed pointer gets incremented by size of the type it points to.
```
Rageking8, line 159: > A `void*` gets incremented by the size of a **`char`** (1 byte).

This implies that pointer arithmetic can be done on `void*`, but that causes a hard error, since it is non-standard and only available as an extension on GCC (not sure if this extension exists for MSVC). Hence, this sentence is removed, but we can consider explicitly stating that it is not allowed.

docs/cpp/raw-pointers.md
```diff
@@ -156,7 +156,11 @@ int main()
...
+```Output
+1 2 3 4 5
```
Rageking8, line 160: There is a trailing space here as that's the verbatim output of this snippet.

ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

docs/cpp/raw-pointers.md
TylerMSFT: excellent clarification

docs/cpp/raw-pointers.md
TylerMSFT: Agree. The article was a fun read.

TylerMSFT:
#sign-off

PR #5626, opened 2025-08-06
Elide some wrong uses of `/c` flag in C runtime library

Rageking8:
All of these examples require linking to function, hence the use of `/c` is invalid and needs to be removed.


ttorble:
@TylerMSFT 

#label:"aq-pr-triaged" 

TylerMSFT:
#sign-off

PR #5731, opened 2025-09-02
Use environment variables to specify compiler preferences.

crud89:
The documentation is using a problematic way to specify the compiler in a preset, namely by setting the `CMAKE_<LANG>_COMPILER` cache variables. This should be avoided, as it can cause problems with subsequent cache configuration runs. An example for where this becomes problematic is described [here](https://discourse.cmake.org/t/unexpected-behavior-when-regenerating-a-project-with-ninja-multi-config-and-msvc/14099).

The [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html) for those variables state: 

> This is the command that will be used as the <LANG> compiler. **Once set, you can not change this variable.**

However, specifying a cache variable in a CMake preset will cause the variable to be reset with every subsequent cache configuration run. This usually results in a warning like the following:

> You have changed variables that require your cache to be deleted.
> Configure will be re-run and you may have to reset some variables.
> The following variables have changed:
> CMAKE_C_COMPILER= cl
> CMAKE_CXX_COMPILER= cl

In the best case, this will simply delete and re-create the entire cache. In the worst case, however, this re-creation will cause problems as previously cached entries are reset. I frequently encountered this problem when resolving packages from vcpkg. The directories to the *Config.cmake files were removed during this process and the actual configuration step then failed because the package could no longer be resolved. This is the same problem as observed in the [linked post above](https://discourse.cmake.org/t/unexpected-behavior-when-regenerating-a-project-with-ninja-multi-config-and-msvc/14099).

Instead of setting those cache variables directly, I've changed the documentation to use the `CXX` and `CC` environment variables. Those are used only during the initial cache configuration run to initialize the `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` variables. This fixes the aforementioned issues in all projects, I've tested it.


v-dirichards:
@TylerMSFT 

#label:"aq-pr-triaged"

TylerMSFT:
#sign-off

PR #5808, opened 2025-09-28
Add/change some code block languages to `cmake`

Rageking8:
Use `cmake` as the code block language for clarity and syntax highlighting as seen on [this page](https://github.com/MicrosoftDocs/vcpkg-docs/blob/2c4d444297b0892c9303b6d4d698821dce165086/vcpkg/users/platforms/windows.md):

image

PR #5814, opened 2025-09-30
Set CMake code block language to `txt` as a fallback

Rageking8:
Since code block language `cmake` is not supported in this repository (https://github.com/MicrosoftDocs/cpp-docs/pull/5808#pullrequestreview-3286050201), we have mainly 3 options:

1. Keep it empty
    - No syntax highlighting
    - Might be seen as "missing" or "incomplete"
1. Put `txt`
    - No syntax highlighting
    - Matches the `.txt` extension of `CMakeLists.txt`
    - Could be mistaken as a normal text file/output rather than CMake
1. Put some other language
    - Potentially plausible syntax highlighting
    - Might be confusing or misleading

Overall, I am in favor of option 2.


TylerMSFT:
#sign-off

